PR #6, opened 2024-10-18
Switch to using langchain.js

jtpio:
Experiments towards https://github.com/jupyterlite/jupyterlab-codestral/issues/5

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -12,9 +12,9 @@ import {
 import { UUID } from '@lumino/coreutils';
 import type { ChatMistralAI } from '@langchain/mistralai';
 import {
+  AIMessage,
   HumanMessage,
-  mergeMessageRuns,
-  SystemMessage
+  mergeMessageRuns
 } from '@langchain/core/messages';
 
 export type ConnectionMessage = {
@@ -39,21 +39,13 @@ export class CodestralHandler extends ChatModel {
     };
     this.messageAdded(msg);
     this._history.messages.push(msg);
-    // const response = await this._mistralClient.chat({
-    //   model: 'codestral-latest',
-    //   messages: this._history.messages.map(msg => {
-    //     return {
-    //       role: msg.sender === 'User' ? 'user' : 'assistant',
-    //       content: msg.body
-    //     };
-    //   })
-    // });
+
     const messages = mergeMessageRuns(
       this._history.messages.map(msg => {
         if (msg.sender.username === 'User') {
           return new HumanMessage(msg.body);
         }
-        return new SystemMessage(msg.body);
+        return new AIMessage(msg.body);
       })
     );
     const response = await this._mistralClient.invoke(messages);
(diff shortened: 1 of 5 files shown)


src/handler.ts
```diff
@@ -34,22 +39,29 @@ export class CodestralHandler extends ChatModel {
...
+    //       content: msg.body
+    //     };
+    //   })
+    // });
+    const messages = mergeMessageRuns(
+      this._history.messages.map(msg => {
+        if (msg.sender.username === 'User') {
+          return new HumanMessage(msg.body);
+        }
+        return new SystemMessage(msg.body);
```
brichet, line 56: Do we need to stack all the previous messages ?

I see that it was already the case before, but I wonder if it is necessary, it can become huge for a long conversation.

jtpio: Yes it can also increase the payload for the requests, but it gives the full context of the conversation as there is otherwise no way to have some kind of "memory"?

But maybe we can try using some higher level primitives offered by `langgraph` to hide this logic: https://js.langchain.com/docs/how_to/message_history/

jtpio: Also it looks like this `SystemMessage` should be switched to a `AIMessage`: https://js.langchain.com/docs/concepts/#aimessage

brichet: > Yes it can also increase the payload for the requests, but it gives the full context of the conversation as there is otherwise no way to have some kind of "memory"?

Right, I was wondering if there could be some kind of "session" when using the API, and this history to be stored on the server providing the API.

> But maybe we can try using some higher level primitives offered by `langgraph` to hide this logic: https://js.langchain.com/docs/how_to/message_history/

We can probably keep it that way for a first PR, and think on how to improve the history in future issues / PRs.

brichet: > Also it looks like this `SystemMessage` should be switched to a `AIMessage`: https://js.langchain.com/docs/concepts/#aimessage

Good catch.
Also the bot messages were not added to the handler history, nor to the API messages.

src/index.ts
```diff
@@ -29,6 +27,10 @@ const inlineProviderPlugin: JupyterFrontEndPlugin<void> = {
     manager: ICompletionProviderManager,
     settingRegistry: ISettingRegistry
   ): void => {
+    const mistralClient = new MistralAI({
```
jtpio, line 30: Also we'll probably want to revisit this later to handle multiple providers:

- do we need to instantiate both a `MistralAI` and `ChatMistralAI` clients?
- maybe there should be one plugin to provide the AI clients, that other plugins would consume
- offer a way to use in set the API key without going through the settings system, since the settings are stored in the browser in JupyterLite

jtpio:
Opened https://github.com/jupyterlite/jupyterlab-codestral/issues/9 as a follow-up, feel free to open other issues to track more improvements.

@brichet if this looks good to you for now, maybe we can undraft and merge, then iterate?

PR #10, opened 2024-10-28
Making the LLM providers more generics

brichet:
This PR adds the possibility of changing LLM provider, even if only *MistralAI* is available at the moment.

A unique *inline completion provider* is registered to the manager.
This *inline completion provider* relies on a completer specific to the LLM used, to correctly call the LLM API.

Adding a new LLM would be to:
- add the completer class in *llm-models* directory
- modify the *llm-models/utils.ts* to return the correct models
- modify the *schema/llm-provider.json* to add the new option

**TODO, probably in a follow up PR**: dynamically update the available settings when the selected provider has changed.


src/chat-handler.ts
```diff
@@ -38,6 +45,19 @@ export class CodestralHandler extends ChatModel {
       type: 'msg'
     };
     this.messageAdded(msg);
+
+    if (this._llmClient === null) {
+      const botMsg: IChatMessage = {
+        id: UUID.uuid4(),
+        body: '**Chat client not configured**',
```
jtpio, line 52: Maybe we can open a new issue to track translating this string (and maybe others) with the `ITranslator`?

src/chat-handler.ts
```diff
@@ -9,23 +9,30 @@ import {
...
-export class CodestralHandler extends ChatModel {
-  constructor(options: CodestralHandler.IOptions) {
+export class ChatHandler extends ChatModel {
+  constructor(options: ChatHandler.IOptions) {
     super(options);
-    this._mistralClient = options.mistralClient;
+    this._llmClient = options.llmClient;
+  }
+
+  get llmClient(): BaseChatModel | null {
```
jtpio, line 31: Wondering if "provider" could also be a good name as an alternative to `llmClient`? Or something else similar to what Jupyter AI uses?

src/index.ts
```diff
@@ -148,4 +94,38 @@ const chatPlugin: JupyterFrontEndPlugin<void> = {
   }
 };
 
-export default [inlineProviderPlugin, chatPlugin];
+const llmProviderPlugin: JupyterFrontEndPlugin<ILlmProvider> = {
+  id: 'jupyterlab-codestral:llm-provider',
+  autoStart: true,
+  requires: [ICompletionProviderManager, ISettingRegistry],
+  provides: ILlmProvider,
```
jtpio, line 101: `ILlmProvider` may read a bit awkward  with the change of casing between the two "L". Maybe `ILLMProvider` or `IAIProvider` could also work? (or something else)

src/chat-handler.ts
brichet: https://github.com/jupyterlite/jupyterlab-codestral/issues/12

jtpio (approved):
Thanks @brichet!

Looks like we can get this one in and iterate more on it.

jtpio:
It would be interesting to follow-up by adding support for another provider, for example OpenAI.

PR #14, opened 2024-11-04
Dynamic settings for providers

brichet:
Create provider settings at build time and update the settings list dynamically.
The JSON schema is created from the typescript interface of the chat provider.

Peek 2024-11-04 16-25

It needs https://github.com/jupyterlab/jupyterlab/pull/16907 to works properly (live update).

It works for Mistral AI, not sure yet about other providers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,9 @@ jupyterlite_ai/labextension
 # Version file is handled by hatchling
 jupyterlite_ai/_version.py
 
+# Settings schema are built
+src/_provider-settings
+
 # Created by https://www.gitignore.io/api/python
 # Edit at https://www.gitignore.io/?templates=python
 
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 AI code completions and chat for JupyterLab, Notebook 7 and JupyterLite, powered by MistralAI ✨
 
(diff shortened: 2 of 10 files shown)


jtpio:
Nice, thanks @brichet!

> It needs [jupyterlab/jupyterlab#16907](https://github.com/jupyterlab/jupyterlab/pull/16907) to works properly (live update).

Looks like this PR will go be available in `4.4.0`, so there will likely be some delay before it is available in JupyterLite.

jtpio:
Since this extension is still experimental anyway, maybe it would be fine to require JupyterLab 4.4.0 / JupyterLite 0.6.0, which are currently in pre-release?

brichet:
> Since this extension is still experimental anyway, maybe it would be fine to require JupyterLab 4.4.0 / JupyterLite 0.6.0, which are currently in pre-release?

You mean adding the python dependencies ? Maybe we could add it as optional dependencies `lab` and `lite` ?

jtpio:
I mean if someone wants to use this extension they will need a version which supports the dynamic settings. 

Maybe just a quick note in the README would be enough?

jtpio (approved):
Thanks!

Let's go with this for now then 👍

PR #15, opened 2024-11-05
Refactoring AIProvider and handling errors

brichet:
This PR is to prepare adding new providers, see https://github.com/jupyterlite/jupyterlab-codestral/issues/13#issuecomment-2456729648 for context.

### Changes

- The chat and completer providers are created every time a setting change, since some of them are not updating the config after init.
- The AIProvider is embedded in the chat handler
- If there is an error in the settings of the chat provider or while sending a message, the error message is displayed as a chat message.
  We may want a better way to display errors (notification, header in the chat...).

Peek 2024-11-05 17-03

### DIFF
--- a/src/chat-handler.ts
+++ b/src/chat-handler.ts
@@ -16,6 +16,8 @@ import {
   mergeMessageRuns
 } from '@langchain/core/messages';
 import { UUID } from '@lumino/coreutils';
+import { getErrorMessage } from './llm-models';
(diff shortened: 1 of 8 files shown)


brichet:
> * We may want a better way to display errors (notification, header in the chat...).

AFAIK, the current implementation of the settings panel does not allow custom validator.
The only way I can see to display a warning / error message in the settings panel is to add it in the DOM directly.

PR #18, opened 2024-11-08
Improves the relevance of codestral completion

### LINKED ISSUE #16: Usability issues with the inline completer
When trying https://github.com/jupyterlite/jupyterlab-codestral/pull/15 locally with MistralAI, the inline completer seems to be "off" most of the time, making it difficult to use:

image

I believe this was already noticed in https://github.com/jupyterlite/jupyterlab-codestral/pull/8 and may be the same issue.

It's not clear yet whether it's because the Mistral API is slow to respond, because of its 1 req/s limit or how the inline completer is currently set up.

brichet:
This PR try to improve the relevance of code completion using codestral.

Should fix #16 

### Context

Codestral allows only one request per second, and a completion request may takes about 2 sec.
This can lead to inconsistency between the suggestion and the current prompt, which may have change during the request.

### Code change

The completion provider keep in memory the latest prompt if the a request is pending.
When the suggestion is received, there are 3 possible situations:
- **the prompt hasn't change**, the suggestion is returned as it
- **the prompt has been updated but still contains the previous one**. For example the fetch may have been requested with `def te` and the current prompt is `def test`.
  - if the suggestion contains the update, the updated part is removed from the suggestion and return it
  - otherwise, a new completion is requested
- **too many differences between the invoked prompt and the current one**, a new completion is requested

The completion may take more time to be display, but should be more relevant.

Additionally, the options `Debouncer delay` in the inline completion settings should be used, to add a delay before fetching the completion after the last key has been pressed.

EDIT: I forgot to mention that it also adds a timeout of 3s, and fetch again if there is no respond in this delay. This is to avoid some infinite waiting from server.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,6 @@
         "@jupyterlab/settingregistry": "^4.2.0",
         "@langchain/core": "^0.3.13",
         "@langchain/mistralai": "^0.1.1",
-        "@lumino/commands": "^2.1.2",
         "@lumino/coreutils": "^2.1.2",
         "@lumino/polling": "^2.1.2",
         "@lumino/signaling": "^2.1.2"
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
         "@jupyterlab/settingregistry": "^4.2.0",
         "@langchain/core": "^0.3.13",
         "@langchain/mistralai": "^0.1.1",
+        "@lumino/commands": "^2.1.2",
         "@lumino/coreutils": "^2.1.2",
         "@lumino/polling": "^2.1.2",
         "@lumino/signaling": "^2.1.2"
--- a/src/completion-provider.ts
+++ b/src/completion-provider.ts
@@ -16,6 +16,7 @@ export class CompletionProvider implements IInlineCompletionProvider {
 
   constructor(options: CompletionProvider.IOptions) {
     const { name, settings } = options;
+    this._requestCompletion = options.requestCompletion;
     this.setCompleter(name, settings);
   }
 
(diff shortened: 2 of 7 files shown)


jtpio:
> **the prompt has been updated but still contains the previous one**. For example the fetch may have been requested with `def te` and the current prompt is `def test`.

Probably this could indeed help, although when typing it's common to go back, delete a few characters, type new ones. So not sure if it really helps in practice?

Also wondering if this may be adding additional complexity to the inline completer logic to work around the slow response time from the Mistral API? Maybe it's already a bit faster with other providers?

brichet:
> Probably this could indeed help, although when typing it's common to go back, delete a few characters, type new ones. So not sure if it really helps in practice?

I agree that it may not often be useful, it only avoid fetching again the completion if the previous one includes the changes...
This part is only an extra of the PR, the main goal is to avoid displaying outdated suggestions, even if it takes time.

> Also wondering if this may be adding additional complexity to the inline completer logic to work around the slow response time from the Mistral API?

The complexity is mostly added to the codestral completer, it should not be ported to other completers if not required.

brichet:
I forgot to mention that it also adds a timeout of 3s, and fetch again if there is no respond in this delay. This is to avoid some infinite waiting from server.

brichet:
@jtpio should we move forward with this PR ?
If you still think that this PR adds complexity, I can remove the part comparing the diff between the current prompt and the suggestion.

The changes from this PR would still avoid displaying outdated suggestions, even if the suggestion can take a long time to be available.

jtpio:
> If you still think that this PR adds complexity, I can remove the part comparing the diff between the current prompt and the suggestion.

Ah yes, maybe it can be a good tradeoff for now for getting that one in.

Also if this change is just for the Mistral completion we could indeed get this in, if it already improves the current behavior.

jtpio (approved):
Thanks!

Let's get this in as this should already improve the current situation.

PR #19, opened 2024-11-08
Add OpenAI provider

brichet:


jtpio:
Just pushed to resolve the `yarn.lock` conflict.

jtpio:
Seems to be working fine locally for both the chat and completer:

https://github.com/user-attachments/assets/70cda7ec-d33c-4afa-9e95-5edf4bfa1f98

For the inline completer, there seems to be some issues similar to the other providers. Maybe there is something to be done to improve the quality of the suggestions in general, by checking the response against the input used when the completer was triggered.

src/llm-models/utils.ts
```diff
@@ -60,20 +66,24 @@ export function getErrorMessage(name: string, error: any): string {
...
+    return error.message;
   }
   return 'Unknown provider';
 }
 
 /*
  * Get an LLM completer from the name.
  */
-export function getSettings(name: string): JSONObject | null {
+export function getSettings(name: string): any {
```
jtpio, line 78: Maybe this `any` is fine for now, but we should maybe check if we can put `JSONObject | null` back as the return type.

brichet:
LGTM, I can't approve it

PR #21, opened 2024-12-03
Update references to the repo after the rename

jtpio:
Follow-up to #20.


brichet (changes requested):
Thanks @jtpio.

I've found a few oversights:
- *package.json* -> description
- *README.md* -> Development install (in the bash comments)

We should also update the screenshot and description about the settings panel in the README, but it can be done in a follow up PR.

jtpio:
Thanks @brichet.

Just fixed those. And yes we should also update the screenshots and description in the usage section of the README. Maybe when we start supporting more providers?

brichet (approved):

Issue #24, opened 2024-12-20
Leverage a Code-tuned SLM in-browser


AABoyles:
### Problem

Requiring the Mistral API key to enable this extension violates the core value offering of Jupyterlite--namely, that it runs in the browser.

### Proposed Solution

Leverage a simple, [in-browser SLM](https://codezen.medium.com/unleashing-the-power-of-self-learning-machines-running-slms-on-your-browser-2ed3f3a3496e) to handle AI code completion. Ideally this should be the default option, providing users with seamless, batteries-included access to the extension's functionality, and extend the option to users to leverage Mistral (or other providers) for faster or higher-quality responses.

jtpio:
Thanks @AABoyles for the suggestion :+1: 

Indeed this extension has a lot of potential. It's still in its early days which is why it was focused on supporting some of the available providers for now.

For reference there is also https://github.com/krassowski/jupyterlab-transformers-completer, which was created after the addition of the inline completer API to JupyterLab. A while back the models were working ok in the browser, but were not really useful for code completion (too small). But maybe the state has improved since.

PR #27, opened 2025-01-24
Add `ChromeAI`

jtpio:
Towards https://github.com/jupyterlite/ai/issues/24

Experimenting with built-in (on device) AI, such as Chrome AI: https://developer.chrome.com/docs/ai/built-in

`langchain.js` supports it as a text completion model, still experimental for now: https://js.langchain.com/docs/integrations/llms/chrome_ai/

This requires:

- using Google Chrome
- enabling some flags
- downloading the model (likely Gemini Nano), which should happen automatically on first use

One of the main advantages (omitting the browser limitation and required flag) is the zero setup needed (no API key).

https://github.com/user-attachments/assets/2043b789-fd00-475e-98db-ebbf3fd76ee0

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,8 @@ Enable the proper flags in Google Chrome.
 
 Then restart Chrome for these changes to take effect.
 
+You can test ChromeAI is enabled in your browser by going to the following URL: https://chromeai.org/
+
 > [!WARNING]
 > On first use, Chrome will download the on-device model, which can be as large as 22GB (according to their docs and at the time of writing).
 > During the download, ChromeAI may not be available via the extension.
--- a/README.md
+++ b/README.md
@@ -62,6 +62,27 @@ The process is different for each provider, so you may refer to their documentat
 
 ![Screenshot showing how to use the chat](./img/3-usage.png)
 
+## Using ChromeAI
+
+> [!WARNING]
+> Support for ChromeAI is still experimental.
+
+Enable the proper flags in Google Chrome.
+
+- chrome://flags/#prompt-api-for-gemini-nano
+- [Optional] chrome://flags/#text-safety-classifier
+
+![a screenshot showing how to enable the ChromeAI flag in Google Chrome](https://github.com/user-attachments/assets/d48f46cc-52ee-4ce5-9eaf-c763cdbee04c)
+
+Then restart Chrome for these changes to take effect.
+
+> [!WARNING]
+> On first use, Chrome will download the on-device model, which can be as large as 22GB (according to their docs).
+> During the download, ChromeAI may not be available via the extension.
+
+> [!NOTE]
+> For more information about Chrome Built-in AI: https://developer.chrome.com/docs/ai/get-started
+
 ## Uninstall
 
 To remove the extension, execute:
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ The process is different for each provider, so you may refer to their documentat
 ## Using ChromeAI
 
 > [!WARNING]
-> Support for ChromeAI is still experimental.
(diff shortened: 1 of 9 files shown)


scripts/settings-generator.js
```diff
@@ -78,6 +82,21 @@ Object.entries(providers).forEach(([name, desc], index) => {
 
   // Remove the properties from extended class.
   const providerKeys = Object.keys(schema.definitions[desc.type]['properties']);
+
+  // TODO: fix for ChromeAI
```
jtpio, line 86: Looks like https://github.com/jupyterlite/ai/pull/19 contains some changes to the `settings-generator` script that be useful here too.

jtpio: Looks like switching to `ChromeAIInputs` should fix it.

jtpio:
OK looks like we could stop here for now as this seems to be working fine, provided that requirements are met (Chrome + flags enabled).

jtpio:
Pushed one more commit to not discard completions that may be formatted as code blocks.

brichet:
Thanks @jtpio.

I had to enable more flags to download the model locally. These steps where describes in https://chromeai.org/ (the link you mentioned in the README. I wonder if we should add these steps directly in the README.

> Necessary Experimental Flags
> 1. chrome://flags/#prompt-api-for-gemini-nano
>   Select 'Enabled'

> 2.  chrome://flags/#optimization-guide-on-device-model
>   Select 'Enabled BypassPrefRequirement'

> 3. chrome://components
>   Click 'Check for Update' on Optimization Guide On Device Model to download the model. If you don't see Optimization Guide, ensure you have set the flags correctly above, relaunch your browser, and refresh the page.

brichet:
Also on my side this is really slow. Do you know if there are some set up to make it faster ?

README.md
```diff
@@ -62,6 +62,29 @@ The process is different for each provider, so you may refer to their documentat
 
 ![Screenshot showing how to use the chat](./img/3-usage.png)
 
+## Using ChromeAI
+
+> [!WARNING]
+> Support for ChromeAI is still experimental and only available in Google Chrome.
```
brichet, line 68: Do you know if there a way to detect the browser, and have this provider available for chrome only ?

jtpio: Maybe just checking for `window.ai` availability could be enough?

jtpio:
> Also on my side this is really slow. Do you know if there are some set up to make it faster ?

Yeah it may be quite slow on some devices indeed.

We could mention the same list of flags as on https://chromeai.org/ for completeness, and also leave a note the performance may not be great on some devices. Not sure there is anything to do about this at the moment.

README.md
jtpio: Since we specify the name of the model as part of the settings definition:

```json
"enum": ["None", "Anthropic", "ChromeAI", "MistralAI"]
```

Not sure there could be a nice way to dynamically enable / disable support for ChromeAI, based on `window.ai`? Unless we manually check for "ChromeAI" and remove it from the settings if `window.ai === undefined`.

brichet: > Not sure there could be a nice way to dynamically enable / disable support for ChromeAI, based on `window.ai`?

Yes, that settings management is already hazardous, don't no neither if we want to add more manual changes.
Maybe the way to go will be to add a dedicated configuration panel that doesn't rely on settings...

brichet (approved):
Thanks, LGTM

README.md
jtpio: > Maybe the way to go will be to add a dedicated configuration panel that doesn't rely on settings...

Agreed, sounds like it would give more flexibility 👍

brichet: I opened https://github.com/jupyterlite/ai/issues/35

Issue #25, opened 2025-01-23
Define a set of default prompts for chat and completions


jtpio:
### Problem

The quality of the responses from the AI providers could likely be improved if we "initialize" the models with a good prompt.

### Proposed Solution

As an example, `jupyter-ai` defines a `CHAT_SYSTEM_PROMPT` here:

https://github.com/jupyterlab/jupyter-ai/blob/becc95ca63c0526bbb3b8e8d040c839ef2efe176/packages/jupyter-ai-magics/jupyter_ai_magics/providers.py#L45-L58

And a `COMPLETION_SYSTEM_PROMPT` here for completions:

https://github.com/jupyterlab/jupyter-ai/blob/becc95ca63c0526bbb3b8e8d040c839ef2efe176/packages/jupyter-ai-magics/jupyter_ai_magics/providers.py#L80-L87

We should likely define similar prompts for the chat and completions.

New prompts could be added later when new features will be added (for example inline fixes).

brichet:
Thanks @jtpio, I'll update it with the same prompts as `jupyter-ai`.

brichet:
I wonder if we want to have these prompts configurable by user, via the setting panel for example.

jtpio:
> I wonder if we want to have these prompts configurable by user, via the setting panel for example.

Yes that could indeed be nice to have.

Issue #9, opened 2024-10-25
Improve the handling of AI clients and API keys


jtpio:
We should look into improving how the clients are instantiated, and the API keys provided:

- do we need to instantiate both a `MistralAI` and `ChatMistralAI` clients?
- maybe there should be one plugin to provide the AI clients, that other plugins would consume
- offer a way to use in set the API key without going through the settings system, since the settings are stored in the browser in JupyterLite

_Originally posted by @jtpio in https://github.com/jupyterlite/jupyterlab-codestral/pull/6#discussion_r1816217804_

brichet:
> * do we need to instantiate both a `MistralAI` and `ChatMistralAI` clients?

They seem to be intended for different uses, respectively completion and chat.

From the API of [ChatMistralAI](https://v03.api.js.langchain.com/classes/_langchain_mistralai.ChatMistralAI.html) and [MistralAI](https://v03.api.js.langchain.com/classes/_langchain_mistralai.MistralAI.html):

- **Class ChatMistralAI<CallOptions>** 
  *Mistral AI chat model integration.*

- **Class MistralAI**
  *MistralAI completions LLM.*

pmkrawczyk:
I am serving Jupyterlite to multiple users who don't have API key. Would there be a way to add config for optional backend API endpoints that would take care of actual communication with LLM providers using a single API key?

PR #28, opened 2025-01-24
Add initial system prompt in ChatHandler and completion

brichet:
Fixes https://github.com/jupyterlite/ai/issues/25

Use the same prompts as `jupyter-ai` (except for the model name, which is not part of the `BaseChatModel` and may not be provided by all chat model).

### NOTES
1. Currently the completion prompt is not used by the codestral completer, because the codestral completer uses a [text completion model](https://js.langchain.com/docs/integrations/llms/), which doesn't seem to allow such prompts.

2. For a follow up PR, we should:

    - use a chat model for codestral instead of LLM (as expressed in https://github.com/jupyterlite/ai/issues/9#issue-2613421821)
      > do we need to instantiate both a MistralAI and ChatMistralAI clients?
    - create a `BaseCompleter` class that may fit with most of the completers

### FOLLOW UP

We could add a setting to let user define their own prompt https://github.com/jupyterlite/ai/issues/25#issuecomment-2615100692

### DIFF
--- a/src/provider.ts
+++ b/src/provider.ts
@@ -8,9 +8,9 @@ import { CompletionProvider } from './completion-provider';
 import { getChatModel, IBaseCompleter } from './llm-models';
 import { IAIProvider } from './token';
 
-export const chatSystemPrompt = (provider_name: string) => `
+export const chatSystemPrompt = (options: AIProvider.IPromptOptions) => `
 You are Jupyternaut, a conversational assistant living in JupyterLab to help users.
-You are not a language model, but rather an application built on a foundation model from ${provider_name}.
+You are not a language model, but rather an application built on a foundation model from ${options.provider_name}.
 You are talkative and you provide lots of specific details from the foundation model's context.
 You may use Markdown to format your response.
 If your response includes code, they must be enclosed in Markdown fenced code blocks (with triple backticks before and after).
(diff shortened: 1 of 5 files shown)


src/provider.ts
```diff
@@ -8,6 +8,30 @@ import { CompletionProvider } from './completion-provider';
 import { getChatModel, IBaseCompleter } from './llm-models';
 import { IAIProvider } from './token';
 
+export const chatSystemPrompt = (provider_name: string) => `
```
jtpio, line 11: Should this function take an option, instead of raw parameters like `provider_name`?

Wondering if this could make it easier to extend later.

jtpio:
> We could add a setting to let user define their own prompt [#25 (comment)](https://github.com/jupyterlite/ai/issues/25#issuecomment-2615100692)

Wondering why we would need a getter / setter on the completer for the `prompt`, as it is currently set to COMPLETION_SYSTEM_PROMPT` by default for now. But this would likely be useful when allowing users to set their own prompts?

brichet:
> > We could add a setting to let user define their own prompt [#25 (comment)](https://github.com/jupyterlite/ai/issues/25#issuecomment-2615100692)
> 
> Wondering why we would need a getter / setter on the completer for the `prompt`, as it is currently set to `COMPLETION_SYSTEM_PROMPT` by default for now. But this would likely be useful when allowing users to set their own prompts?

Yes, and even with the settings it could only be a writable property, since there is nothing more in the getter / setter.

jtpio (approved):
Thanks!

PR #30, opened 2025-01-28
Upgrade the jupyterlite-core package in deployment

brichet:
Upgrade the version of `jupyterlite-core` to the pre-released one in the deployment (https://jupyterlite.github.io/ai/lab/index.html), to allow dynamic settings.

Otherwise, the page has to be reloaded to fill the required `API key` settings.

Peek 2025-01-28 14-43

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -20,7 +20,8 @@ jobs:
           python-version: '3.11'
       - name: Install the dependencies
         run: |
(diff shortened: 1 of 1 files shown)


martinRenou (approved):
LGTM 😄

brichet:
Thanks for the review @martinRenou 

> well I would have used jupyterlite-xeus. But let's agree to disagree on the kernel choice here :P

You should open an issue to discuss it :smile:

PR #31, opened 2025-01-28
Use a chat model instead of LLM for codestral completion

brichet:
`Langchainjs` warns against using the text completion model, and advises using the chat model instead, which may be more up-to-date (see caution at https://js.langchain.com/docs/integrations/llms/).

This PR replace the text completion model by a chat model in codestral completion.

### DIFF
--- a/src/llm-models/codestral-completer.ts
+++ b/src/llm-models/codestral-completer.ts
@@ -19,22 +19,6 @@ import { COMPLETION_SYSTEM_PROMPT } from '../provider';
  */
 const INTERVAL = 1000;
 
-/**
- * The default prompt for the completion request
- * TODO: move somewhere else so it can be used by other completers
- * See: https://github.com/jupyterlite/ai/issues/25
- */
-const DEFAULT_PROMPT = `
-You are an application built to provide helpful code completion suggestions. \
-You should only produce code. Keep comments to minimum, use the \
-programming language comment syntax. Produce clean code. \
-The output should be a single string, and should correspond to what a human users \
-would write after the content of the message, without fenced code block and backtick. \
-The code is written in JupyterLab, a data analysis and code development \
-environment which can execute code extended with additional syntax for \
-interactive features, such as magics.
-`;
-
 export class CodestralCompleter implements IBaseCompleter {
   constructor(options: BaseCompleter.IOptions) {
     this._mistralProvider = new ChatMistralAI({ ...options.settings });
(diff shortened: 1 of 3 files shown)


brichet:
The completion works better when using a `MistralAI` model designed for code completion, like `codestral-latest`.

src/llm-models/codestral-completer.ts
```diff
@@ -15,99 +19,73 @@ import { BaseCompleter, IBaseCompleter } from './base-completer';
 const INTERVAL = 1000;
 
 /**
- * Timeout to avoid endless requests
+ * The default prompt for the completion request
+ * TODO: move somewhere else so it can be used by other completers
+ * See: https://github.com/jupyterlite/ai/issues/25
  */
-const REQUEST_TIMEOUT = 3000;
+const DEFAULT_PROMPT = `
```
jtpio, line 26: Looks like we can now update this PR to use the default prompts from https://github.com/jupyterlite/ai/pull/28?

jtpio:
Testing this locally, completions appear to be quite fast, although the response often includes code block delimiters:

image

Over on https://github.com/jupyterlite/ai/pull/27 there is some logic to not discard responses that may include the code block delimiters, but instead process the string to keep the suggestion without the backticks and language name.

So maybe we could isolate that to some util functions so it can be reused across providers.

jtpio:
We can also improve the default completion prompt to mention that responses should be formatted without code block delimiters.

brichet:
> We can also improve the default completion prompt to mention that responses should be formatted without code block delimiters.

I did some tests with these kind of prompts and it doesn't work, looks like it doesn't care of these instructions.
But it worked better on my side with the `codestral-latest` model, is it the one you tested ?

jtpio:
> But it worked better on my side with the `codestral-latest` model, is it the one you tested ?

Yeah strangely it seems to be returning code blocks quite often?


https://github.com/user-attachments/assets/d86b0924-5410-4a91-b6fd-d034c220aefa

brichet:
> Playing around with it a bit more locally, adding more instructions to the prompt like this one seems to be helping a bit

Thanks for digging into it. Feel free to push in the branch with your changes. Otherwise I can do it.

jtpio:
I pushed a change to improve the default prompt used for completions, so it can also be useful for other providers.

jtpio:
@brichet if this looks good to you, happy to merge as it is right now.

brichet:
> @brichet if this looks good to you, happy to merge as it is right now.

Let's merge it, thank you for the prompt update. 

I still have some unexpected suggestions from MistralAI , but it may be more an issue with the model.

[Peek 2025-02-10 11-51.webm](https://github.com/user-attachments/assets/3280b2bd-5da0-474c-95ea-7475f49ab173)

PR #34, opened 2025-02-07
Magic wand support

jtpio:
Investigating how `@jupyterlite/ai` could make use of the `jupyterlab-magic-wand` extension: https://github.com/Zsailer/jupyterlab-magic-wand

See https://github.com/Zsailer/jupyterlab-magic-wand/issues/4 for more context.

This PR is currently based on https://github.com/Zsailer/jupyterlab-magic-wand/pull/5, to allow swapping the `jupyterlab_magic_wand:magic-provider` with another implementation.

https://github.com/user-attachments/assets/942e97d2-a3ca-4243-9f72-a5b77c1dba50

Unfortunately the diff below the input cell does not work out of the box in the browser. This is because `jupyterlab-cell-diff` expects a diff with the `nbdime` format, which for now requires using the `nbdime` server extension: https://github.com/Zsailer/jupyterlab-magic-wand/issues/4#issuecomment-2642458707

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,8 @@
         "@langchain/mistralai": "^0.1.1",
         "@lumino/coreutils": "^2.1.2",
         "@lumino/polling": "^2.1.2",
-        "@lumino/signaling": "^2.1.2"
+        "@lumino/signaling": "^2.1.2",
+        "jupyterlab_magic_wand": "*"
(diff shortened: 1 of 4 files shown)


jtpio:
Opening as a draft just to share early progress on that front.

PR #39, opened 2025-02-13
Stream responses

### LINKED ISSUE #36: Support streaming responses
### Problem

Currently the AI provider replies in one chunk:

https://github.com/user-attachments/assets/02a3a6c9-e384-46ba-ba2c-006e7d1b40bf

### Proposed Solution

Implement streaming responses, so it looks more natural and akin to other chat interfaces.

jtpio:
Fixes #36 

https://github.com/user-attachments/assets/9fabe827-f7dc-409f-9ba3-62a35538a84a

The tokens do seem to be correctly streamed when looking at the request in the dev tools:

image

For reference langchain handles "auto streaming": https://js.langchain.com/docs/concepts/streaming/#auto-streaming-chat-models

### DIFF
--- a/src/chat-handler.ts
+++ b/src/chat-handler.ts
@@ -88,38 +88,42 @@ export class ChatHandler extends ChatModel {
     );
 
     this.updateWriters([{ username: 'AI' }]);
-    return this._aiProvider.chatModel
-      .invoke(messages)
-      .then(response => {
-        // Some providers may use text-completion models (for example ChromeAI),
-        // for which the response is a string.
-        const content = response.content ?? response;
-        const botMsg: IChatMessage = {
-          id: UUID.uuid4(),
-          body: content.toString(),
-          sender: { username: 'AI' },
-          time: Date.now(),
-          type: 'msg'
-        };
+
+    // create an empty message to be filled by the AI provider
+    const botMsg: IChatMessage = {
+      id: UUID.uuid4(),
+      body: '',
+      sender: { username: 'AI' },
+      time: Date.now(),
+      type: 'msg'
+    };
+
+    let content = '';
+
+    try {
+      for await (const chunk of await this._aiProvider.chatModel.stream(
+        messages
+      )) {
+        content += chunk.content;
+        botMsg.body = content;
         this.messageAdded(botMsg);
-        this._history.messages.push(botMsg);
-        return true;
-      })
-      .catch(reason => {
-        const error = getErrorMessage(this._aiProvider.name, reason);
-        const errorMsg: IChatMessage = {
-          id: UUID.uuid4(),
-          body: `**${error}**`,
-          sender: { username: 'ERROR' },
-          time: Date.now(),
-          type: 'msg'
-        };
-        this.messageAdded(errorMsg);
-        return false;
-      })
-      .finally(() => {
-        this.updateWriters([]);
-      });
+      }
+      this._history.messages.push(botMsg);
+      return true;
+    } catch (reason) {
+      const error = getErrorMessage(this._aiProvider.name, reason);
+      const errorMsg: IChatMessage = {
+        id: UUID.uuid4(),
+        body: `**${error}**`,
+        sender: { username: 'ERROR' },
+        time: Date.now(),
+        type: 'msg'
+      };
+      this.messageAdded(errorMsg);
+      return false;
+    } finally {
+      this.updateWriters([]);
+    }
   }
 
   async getHistory(): Promise<IChatHistory> {
--- a/src/chat-handler.ts
+++ b/src/chat-handler.ts
@@ -87,13 +87,14 @@ export class ChatHandler extends ChatModel {
       })
     );
 
-    this.updateWriters([{ username: 'AI' }]);
+    const sender = { username: 'AI' };
+    this.updateWriters([sender]);
 
     // create an empty message to be filled by the AI provider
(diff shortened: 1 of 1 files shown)


src/chat-handler.ts
```diff
@@ -88,38 +88,42 @@ export class ChatHandler extends ChatModel {
...
+    } catch (reason) {
+      const error = getErrorMessage(this._aiProvider.name, reason);
+      const errorMsg: IChatMessage = {
+        id: UUID.uuid4(),
+        body: `**${error}**`,
+        sender: { username: 'ERROR' },
+        time: Date.now(),
+        type: 'msg'
+      };
+      this.messageAdded(errorMsg);
```
jtpio, line 122: This seems to be doing the job, although there may be some future optimizations to be done in jupyter-chat.

Currently `jupyter-chat` seems to be removing the previous message before adding the one with the update here:

https://github.com/jupyterlab/jupyter-chat/blob/73512c16a6c602b81f873ecf9a3c20cb73cefd94/packages/jupyter-chat/src/model.ts#L435-L449

Maybe it should also compare the timestamps like the comment suggests?

jtpio:
Do we want to let users choose between streaming and not streaming responses?

Probably it's likely fine to default to streaming for now if we can, since this is what all AI chats are doing at the moment. But if we want to let users choose there could be a new option in the settings.

src/chat-handler.ts
brichet: I think the comment do not refer to the correct line, the message is sent/updated above:
```
this.messageAdded(botMsg);
```

In your implementation, the timestamp is not updated, so the position of the message should not change. And it is probably better like this, to avoid confusion.
For comparison, streaming in jupyter-ai lead to [update_message](https://github.com/jupyterlab/jupyter-chat/blob/73512c16a6c602b81f873ecf9a3c20cb73cefd94/python/jupyterlab-chat/jupyterlab_chat/ychat.py#L142-L150) in the back-end, which do not update the timestamp.

We should probably homogenize this in `jupyter-chat`, and allow consumer extension to change or not the timestamp...

jtpio: Yes right it was about the line above (which Github shows as not modified).

> the timestamp is not updated, so the position of the message should not change. And it is probably better like this, to avoid confusion.

Yes, in terms of UX this seems fine.

src/chat-handler.ts
```diff
@@ -87,39 +87,44 @@ export class ChatHandler extends ChatModel {
...
+      type: 'msg'
+    };
+
+    let content = '';
+
+    try {
+      for await (const chunk of await this._aiProvider.chatModel.stream(
+        messages
+      )) {
+        content += chunk.content;
```
brichet, line 108: Wondering if the stream are always additional or could try to modify previous part of the message.
Don't know if there is a setting for that, and if it worth looking for that feature in future.

brichet (approved):
Looks great, thanks @jtpio

src/chat-handler.ts
jtpio: Also on another note for a different issue, after using the `jupyter-chat` API, I was wondering if changing to the following could make sense:

- `messageAdded` -> `addMessage`
- `messagesDeleted` -> `deleteMessage`

In general when working with extensions, when properties finish with `ed` they tend to refer to signals. Also it may more of a personal opinion but `addMessage` sounds more natural than `messageAdded` when interacting with the chat model.

brichet: I agree, but in this case it could also be confusing with `sendMessage`  and `deleteMessage` that already exist.

This is the distinction between:
- the actions the user do in its chat (`sendMessage`  and `deleteMessage`) that must be propagated to the other users
- the actions done by remote users (`messageAdded` and `messagesDeleted`) and that must update the chat view...

In the case of `jupyterlite` chats, it is most confusing because there is no back-end, the `sendMessage` propagates itself the messages from the user and the AI.

It may worth an issue in `jupyter-chat` to discuss the naming.

src/chat-handler.ts
jtpio: Guessing they should be additional as otherwise it would be difficult to tell if that's the case or not?

jtpio: Or you mean if some providers would return the complete updated message each time?

For example:

- He
- Hello
- Hello,
- Hello, world

Instead of:

- He
- llo
- ,
- world

The second one is more efficient in terms of bandwidth and data being transferred, so there may not be many advantages for the first approach?

src/chat-handler.ts
jtpio: Yeah happy to open an issue on https://github.com/jupyterlab/jupyter-chat to discuss this 👍

jtpio:
Thanks @brichet for the review!

Merging to be able to use this change in other PRs. Happy to track questions and improvements separately.

src/chat-handler.ts
brichet: In this case I agree that additional is the best approach.

> Guessing they should be additional as otherwise it would be difficult to tell if that's the case or not?

Indeed, it must always act in the same way.

To explain my question, it looks like it's writing word by word, like a human. But sometime you realise in the middle of the sentence that you could have wrote it differently for more consistency, and want to reformat the whole sentence. I was wondering if some model could act similarly to human. But as you said, it would be difficult to know with this API.

PR #40, opened 2025-02-13
Add icon and name for the AI assistant

jtpio:
- [x] Add an avatar for the AI assistant, based on the one from Jupyter AI, but using the JupyterLite colors:
- [x] Add `personaName` to the chat settings, defaulting to `Jupyternaut` for now

image


src/chat-handler.ts
```diff
@@ -87,7 +104,9 @@ export class ChatHandler extends ChatModel {
       })
     );
 
-    this.updateWriters([{ username: 'AI' }]);
+    const avatar = `data:image/svg+xml;base64,${AI_AVATAR_BASE64}`;
```
jtpio, line 107: Maybe `jupyter-chat` could allow taking a `LabIcon` as `avatar_url`, or expose an `avatar` field which could be a `LabIcon`.

brichet: This interface is a copy of [`User.IIdentity`](https://github.com/jupyterlab/jupyterlab/blob/2f674b04bf33c46da561aa5b74dbc92aed2777c7/packages/services/src/user/index.ts#L235), that is compatible with the [server version of it](https://github.com/jupyter-server/jupyter_server/blob/641e8fccad2e96ba305a74ce83e9420083b7fca6/jupyter_server/auth/identity.py#L35).

In `jupyterlab-chat` extension, the user data are saved in the collaborative document (chat file) in plain text. I don't know if there would be a way to keep the `LabIcon` in this case (some kind of serialization ?)

brichet (commented):
Really nice, thanks @jtpio.

src/chat-handler.ts
```diff
@@ -42,6 +49,16 @@ export class ChatHandler extends ChatModel {
...
 
+  /**
+   * Getter and setter for the persona name.
+   */
+  get personaName(): string {
+    return this._personaName;
+  }
+  set personaName(value: string) {
+    this._personaName = value;
+  }
```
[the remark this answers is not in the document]
jtpio: That sounds like a good idea yes 👍

src/chat-handler.ts
jtpio: Right, not sure either for now (haven't investigated more), but sticking to this approach is likely fine for now.

jtpio:
Pushed one more change to use the jupyter chat icon in the settings:

image

brichet (approved):
Thanks @jtpio

PR #41, opened 2025-02-13
Add chat autocompletion with the `/clear` command

jtpio:
Slash commands are useful to perform regular actions, such as clearing the history of messages.

This adds a `IAutocompletionRegistry` to register `/clear` and be able to clear the chat history.

https://github.com/user-attachments/assets/fce46181-e722-43c0-b320-eaebf8237ed5


jtpio:
We may also need to apply the same trick as in https://github.com/jupyterlab/jupyter-chat/pull/159.

brichet:
Thanks for looking into it @jtpio.

FYI there may be some changes in the way the autocompletion are handled and registered in jupyter-chat:
- https://github.com/jupyterlab/jupyter-chat/issues/144
- https://github.com/jupyterlab/jupyter-chat/pull/161

jtpio:
> FYI there may be some changes in the way the autocompletion are handled and registered in jupyter-chat:

Thanks for the heads-up. Sounds good, we can adapt here after updating to a new version of `jupyter-chat`.

src/chat-handler.ts
```diff
@@ -78,10 +78,16 @@ export class ChatHandler extends ChatModel {
   }
 
   async sendMessage(message: INewMessage): Promise<boolean> {
+    const body = message.body;
+    if (body.startsWith('/clear')) {
+      // TODO: do we need a clear method?
+      this.messagesDeleted(0, this.messages.length);
+      return false;
+    }
```
brichet, line 86: We should also clear the message history `this._history.messages`, to avoid sending cleared messages.

brichet (changes requested):

brichet:
There seems to be an issue with keyboard when displaying and hiding the completer suggestion.
In the example below, pressing enter should send the message, but does not work anymore if the autocomplion has been opened 

[record-2025-02-14_11.13.18.webm](https://github.com/user-attachments/assets/e3ce62e2-b185-4d8e-85ef-9db03fc02005)

I think that it is an issue with `jupyter-chat`, I will make some tests there.

brichet:
> We should also clear the message history this._history.messages, to avoid sending cleared messages.

This make me realize that the history seems to never be cleared, which could lead to a huge message list. I'll open an issue on this.

jtpio:
> In the example below, pressing enter should send the message, but does not work anymore if the autocomplion has been opened

It also seems to be an issue with the Jupyter AI v3 pre-release. For example when testing from the `main` branch of https://github.com/jupyterlab/jupyter-ai:


https://github.com/user-attachments/assets/204a7002-bb17-4f2e-a459-7c2e979074a3

brichet (approved):
Thanks @jtpio, looks great.

I also tested the linting error on MUI icons, it works well :+1:

PR #43, opened 2025-02-15
Update to a newer `@langchain/community` to fix ChromeAI integration

jtpio:
ChromeAI was recently giving some really weird results, pretty much gibberish.

This was odd because using the API in the dev tools directly seems to be working fine:

image

Digging a bit more into it, it looks like an issue in the `@langchain/community` package.

I opened a PR upstream, which seems to be fixing the issue with the latest Chrome 133 release (as of today): https://github.com/langchain-ai/langchainjs/pull/7700

**Before**

image

**After**

image

With streaming:

https://github.com/user-attachments/assets/b6706f7f-6537-497c-8206-1c9c56f090ba


patches/@langchain+community+0.3.30.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/@langchain/community/dist/experimental/llms/chrome_ai.js b/node_modules/@langchain/community/dist/experimental/llms/chrome_ai.js
```
jtpio, line 1: This (and the `patch-package` / `postinstall-postinstall` dependencies) can be removed later after updating to a newer version of `@langchain/community` what would include https://github.com/langchain-ai/langchainjs/pull/7700.

brichet:
Thanks @jtpio.
I was wondering if it was a recent change in `chrome-ai` or in `@langchainjs` but just realize that this is following the default use of streaming in this repo, after https://github.com/jupyterlite/ai/pull/39

jtpio:
> I was wondering if it was a recent change in `chrome-ai` or in `@langchainjs` but just realize that this is following the default use of streaming in this repo, after #39

Right, but I also tried locally with the wheel built by check-release in https://github.com/jupyterlite/ai/pull/27 (before supporting streaming), and it was the same issue.

jtpio:
https://github.com/langchain-ai/langchainjs/pull/7700 has been merged and released, so we can remove the patch here.

jtpio:
Thanks for the review!

> BTW, I didn't know the patch workaround, nice to temporary fix an upstream issue 👍

Yeah it is indeed convenient and somehow "clean" :)

PR #47, opened 2025-02-17
WebLLM

jtpio:
Looking into supporting https://js.langchain.com/docs/integrations/chat/web_llm/.

Towards #24 

Not fully functionally yet, but some progress already.

- [x] Basic support for `WebLLM` models in the chat
- [x] Support for the completer?
- [x] Fix TypeScript build issue (remove `skipLibCheck`)
- [x] Document limitations (browser, hardware, performance, bandwidth)
- [x] Document which models are available: https://github.com/mlc-ai/web-llm/blob/632d34725629b480b5b2772379ef5c150b1286f0/src/config.ts#L303-L309

The nice thing is that once a model has been downloaded, the next time it needs to be loaded will be much faster since it's stored in the browser cache:

https://github.com/user-attachments/assets/845999c8-4951-480d-bcb6-0453be247e16


jtpio:
The completer does not work super well, and it seems to be working slightly better with a small model, for example `Qwen2.5-Coder-0.5B-Instruct-q4f32_1-MLC`.

jtpio:
Maybe we can stop here for now.

Support is still experimental, but it can still be useful to some people and also for demos.

jtpio:
@brichet this one may be trickier to test as it requires WebGPU. Maybe using a smaller model like `Qwen2.5-Coder-0.5B-Instruct-q4f32_1-MLC` can make testing easier.

Otherwise maybe we could finally get this one in, so it can more easily be tested and demoed without having to rebase the PR each time.

Maybe we can also look into restructuring the repo a bit later if we start adding more providers that run in the browser, for example https://github.com/huggingface/transformers.js. These may for now not still be at the level of online providers like MistralAI or Anthropic, so they could also be marked as experimental.

brichet:
Indeed, I'm not able to test it now, but let's merge it if it works for you.

brichet:
Also we should be able to detect if the web browser is compatible with WebGPU, and either add a warning in the settings panel or disable the provider if it is not available.

On my side it is stuck to `Loading model...` without error in UI (only in console).

Issue #45, opened 2025-02-17
Allow custom providers via third-party extensions


jtpio:
### Problem

Just like Jupyter AI allows for custom providers: https://jupyter-ai.readthedocs.io/en/latest/developers/index.html#custom-model-providers

It would be great to let folks add other providers, using the JupyterLab extension mechanism.

### Proposed Solution

Likely to be done after #35.

Provide a way to register a new provider with capabilities (chat / completion).

### Additional context

Maybe after #35 it should also show all the registered providers (core + from extensions).

brichet:
We could probably have a registry that should replace of the [`utils` module](https://github.com/jupyterlite/ai/blob/main/src/llm-models/utils.ts), to set/get the providers.

If I'm not mistaken, each completer should provides:
- the chat model
- the completer class
- the settings schema
- optionally the error handler, which can be different depending on the provider (https://github.com/jupyterlite/ai/blob/43aa54c92f1a75a834a2f348f0502e89a0f66764/src/llm-models/utils.ts#L78)
- optionally the instructions to be displayed in the settings

In any case, it would be a nice improvement form the current state, that would clarify/centralize all the provider's accessors.

Issue #37, opened 2025-02-10
Secrets manager


jtpio:
### Problem

Currently API keys are stored via the settings system, which ends up in the browser storage in the case of JupyterLite and on the server for JupyterLab.

This is not great, because it can easily be retrieved, and is also stored in plain text.

### Proposed Solution

Have a proper way to store API keys and secrets.

The default implementation could just store them in memory for the duration of the session.

Ultimately this should be pluggable, so folks can provide different implementations for example if they want to store and retrieve secrets from a remote location.

### Additional context

Likely to be done after:

- https://github.com/jupyterlite/ai/issues/35

If something similar ends up in Jupyter AI, replace the plugin providing that functionality:

- https://github.com/jupyterlab/jupyter-ai/issues/1237
- https://github.com/frankzickert/jupyterlab-credentialstore

jtpio:
There is also the question of whether secrets should be managed with a different panel than the one that will be provided by https://github.com/jupyterlite/ai/issues/35.

While it makes sense to have a generic secrets manager UI, in the context of `@jupyterlite/ai` and Jupyter AI, being able to set the API keys next to other model parameters makes sense too.

brichet:
Cross reference to #53 which make use of a secrets manager in the settings.

> Currently API keys are stored via the settings system, which ends up in the browser storage in the case of JupyterLite and on the server for JupyterLab.

In the current state, secrets are still saved in the settings, because the `AIProvider` update the current provider using the settings:
https://github.com/jupyterlite/ai/blob/2c7dc9e336f9e223644a4ed6a9cd897297795073/src/index.ts#L164-L167

Maybe we could:
- hide the values of the secrets in the settings
- pass the secrets manager to the `AIProvider` object
- let him retrieve the secrets when [setting the provider](https://github.com/jupyterlite/ai/blob/2c7dc9e336f9e223644a4ed6a9cd897297795073/src/provider.ts#L137)

jtpio:
Yes this PR is quite recent but should be available in the latest JupyterLab 4.4.0b1 now. Since we are already using pre-release packages we could try to do the update.

PR #48, opened 2025-02-19
Settings UI improvement

### LINKED ISSUE #35: Add a dedicated settings panel
### Problem

Each provider has its own settings in `@langchainjs`.
Therefore, we need to [update manually](https://github.com/jupyterlite/ai/blob/b5358b941487672850a7c118e552d19add70d605/src/index.ts#L115) the settings when changing the provider.

### Proposed Solution

Have a dedicated AI settings panel that does not use the Jupyterlab settings system, to give more flexibility.

### Additional context

Discussion started at https://github.com/jupyterlite/ai/pull/27#discussion_r1946366872

brichet:
Fixes #35 

[record-2025-02-20_18.17.07.webm](https://github.com/user-attachments/assets/51bb4ccd-3b42-41b8-9d37-28647f2f5856)

This PR uses a react component in the settings panel, to make it more reactive and configurable.
With this new setting component:
- settings are saved for each provider, and restored when loading the page again (use local storage)
- the field containing the word 'key' are replaced by a password widget

It also use a simpler way to import generated schema, no need to set it up manually when adding a new provider.


brichet:
This PR is missing some helpers about the use of each provider, as described in https://github.com/jupyterlite/ai/issues/35#issuecomment-2661519889.

Since this is using a `FormComponent`, helper could easily be included above or below it.
A better user experience could be to include it between the provider select and the dynamic settings. This option needs some refactoring to split the form, and include the helper.

brichet:
Instructions added to the settings (currently only for `MistralAI` and `ChromeAI`), those from the readme.

[record-2025-02-27_13.15.05.webm](https://github.com/user-attachments/assets/c7e990b2-1f59-491e-8b66-6a0fdd926044)

jtpio:
> I explored the use of the `formContext` props, but it doesn't seems that we can pass additional properties to it currently (see [here](https://github.com/jupyterlab/jupyterlab/blob/1c64e5e00c3842eb7694190ed80a31685dcdeffa/packages/settingeditor/src/SettingsFormEditor.tsx#L119-L123))

Thanks for looking into this 👍 

Having a way to provide more parameters would indeed also be useful to pass a secret manager (https://github.com/jupyterlite/ai/issues/37), so the secrets could be saved elsewhere.

For now I guess we can use the trick of the static method, but maybe there is something so be done in JupyterLab to make this more flexible.

src/settings/instructions.ts
```diff
@@ -0,0 +1,50 @@
+export interface IDict<T = any> {
+  [key: string]: T;
+}
+
+const chromeAiInstructions = `
```
jtpio, line 5: Probably fine to leave them here in one file here for now.

Later when third-party extensions can register additional AI providers (https://github.com/jupyterlite/ai/issues/45) they could also provide their own instructions for the provider.

src/settings/panel.tsx
```diff
@@ -0,0 +1,218 @@
...
+      data: { [MD_MIME_TYPE]: mdStr }
+    });
+    await renderer.renderModel(model);
+    this.setState({ instruction: renderer.node });
+  }
+  /**
+   * Get the current provider from the local storage.
+   */
+  getCurrentProvider(): string {
+    const settings = JSON.parse(localStorage.getItem(STORAGE_NAME) || '{}');
```
jtpio, line 87: Wondering how this would play with a custom `overrides.json`, in case some deployments would like to pre-configure a model (and some of its options) by default?

brichet: That would probably overwrite it.
I don't really know how works `overrides.json`. Can we get its default values with [`annotatedDefaults()`](https://github.com/jupyterlab/jupyterlab/blob/1fb48d636f90928c3b4d883d80030f9bc6da6f73/packages/settingregistry/src/tokens.ts#L529) ?
In this case we could simply replace `'{}'` with `this._settingsRegistry.annotatedDefaults()`.

BTW, we should probably rename `this._settingsRegistry` to `this.settings` for consistency.

brichet: Looks like `annotatedDefaults()` returns the `overrides.json` contents.

In the constructor, we could check if there is a value in `localStorage.getItem(STORAGE_NAME)`.
If not, read the `annotatedDefaults()` and, if the provider is set in the default settings, fill the local storage with its contents and set the default provider as current. If the provider is not set in the default settings we can probably not use these values.

brichet:
@jtpio this PR is ready.

- the AISettings component does not use a static rendermime registry anymore. The renderer uses a factory to provide the rendermime registry to the component (thanks again @jtpio for providing this solution)
- if there are no settings in the local storage, it uses the user settings (which lead to the default settings if empty too)

src/settings/panel.tsx
```diff
@@ -0,0 +1,245 @@
...
+
+    this._renderInstruction();
+
+    // Update the setting registry.
+    this._settings
+      .set('AIprovider', this._currentSettings)
+      .catch(console.error);
+  }
+
+  async _renderInstruction(): Promise<void> {
```
jtpio, line 94: Maybe this method can be kept private for now?

```suggestion
  private async  _renderInstruction(): Promise<void> {
```

src/settings/panel.tsx
```diff
@@ -0,0 +1,245 @@
...
+import { JSONSchema7 } from 'json-schema';
+import React from 'react';
+
+import { IDict, instructions } from './instructions';
+import baseSettings from './schemas/base.json';
+import ProviderSettings from './schemas';
+
+const MD_MIME_TYPE = 'text/markdown';
+const STORAGE_NAME = '@jupyterlite/ai:settings';
+const INSTRUCTION_CLASS = 'jp-lite-ai-settings-instruction';
```
jtpio, line 17: If we want to follow a similar naming style as in JupyterLab, maybe the following class name could also work?

```suggestion
const INSTRUCTION_CLASS = 'jp-AISettingsInstructions';
```

jtpio (approved):
Thanks!

PR #49, opened 2025-02-28
Completer plugin

brichet:
This PR add a new plugin for the completion provider, for more consistency with the chat plugin.

Before, the completion provider was embedded in the `AIProvider`, whereas the chat was instantiated in a plugin, and was listening for changes in the AIProvider.

Now, the completion provider is also instantiating in a plugin, and is also listening for change in the AIProvider.
This add more flexibility to disable one of the chat or completer.

It also rename some attributes/function in the `AIProvider`, from model to provider to avoid confusion with the model name of each LLM provider.

### DIFF
--- a/src/chat-handler.ts
+++ b/src/chat-handler.ts
@@ -40,7 +40,7 @@ export class ChatHandler extends ChatModel {
     this._aiProvider = options.aiProvider;
     this._prompt = chatSystemPrompt({ provider_name: this._aiProvider.name });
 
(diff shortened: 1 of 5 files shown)


brichet:
This is a first step related to https://github.com/jupyterlite/ai/issues/45.
With `AIProvider` only used to provide the current completer and chat provider, it will be easier to switch to a registry.

jtpio (approved):
Thanks!

PR #50, opened 2025-03-03
Provider registry

brichet:


src/token.ts
jtpio, line 1: Maybe we can take the opportunity to rename the file to `tokens.ts` (plural)?

src/token.ts
```diff
@@ -1,19 +1,116 @@
...
+   */
+  readonly providers: string[];
+  /**
+   * Add a new provider.
+   */
+  add(provider: IAIProvider): void;
+  /**
+   * Remove a provider.
+   */
+  remove(name: string): void;
```
jtpio, line 63: Wondering if we should expose a `remove` function?

Keeping registries "additive" would help prevent other extensions from removing existing providers.

brichet: So we should also check if the provider already exist before setting it, to make it "additive" only.

jtpio: Right, so there is only one provider with the same name.

If an extension would like to replace an existing provider, it could disable the plugin providing it and add it in its own plugin.

brichet: Done

jtpio (approved):
Thanks!

PR #53, opened 2025-03-06
Use the secrets manager

brichet:
This PR uses the secrets manager to handle passwords in settings.

~~The tests are not intended to work, since the secrets manager package has not been published yet.
To test it, one should install locally [jupyter-secrets-manager](https://github.com/jupyterlab-contrib/jupyter-secrets-manager) and use [yalc](https://github.com/wclr/yalc) to link the Javascript package.~~

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -17,6 +17,7 @@ import { IRenderMimeRegistry } from '@jupyterlab/rendermime';
 import { ISettingRegistry } from '@jupyterlab/settingregistry';
 import { IFormRendererRegistry } from '@jupyterlab/ui-components';
 import { ReadonlyPartialJSONObject } from '@lumino/coreutils';
+import { ISecretsManager } from 'jupyter-secrets-manager';
 
 import { ChatHandler } from './chat-handler';
 import { CompletionProvider } from './completion-provider';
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -137,19 +138,20 @@ const providerRegistryPlugin: JupyterFrontEndPlugin<IAIProviderRegistry> = {
   id: '@jupyterlite/ai:provider-registry',
   autoStart: true,
   requires: [IFormRendererRegistry, ISettingRegistry],
-  optional: [IRenderMimeRegistry],
+  optional: [IRenderMimeRegistry, ISecretsManager],
```
jtpio, line 141: Looks like with this change, API keys are not stored on the page anymore since it will be using the default in-memory secret manager?

Wondering if we should still have a secret manager that stores secrets in the browser, for convenience? Or maybe have an option to choose between in-memory and in-browser storage?

brichet: > Looks like with this change, API keys are not stored on the page anymore since it will be using the default in-memory secret manager?

Right

> Wondering if we should still have a secret manager that stores secrets in the browser, for convenience?

There is one available in the secrets manager repository: [LocalStorageConnector](https://github.com/jupyterlab-contrib/jupyter-secrets-manager/blob/main/src/connectors/local-storage.ts).

But the secrets manager does not allow change of connector on the fly, and that's probably better for security reason. This means that the extension need to choose one or the other at startup I guess.

> Or maybe have an option to choose between in-memory and in-browser storage?

It is probably the simplest solution, a toggle button whether to use or not the secrets manager. If it is not used, the secrets would be stored in the local storage as any other settings.

brichet: There is now an option to disable the secrets manager.

- when the secrets manager is disabled, (1) the current form is saved in the local storage, to include the passwords that was previously not saved, and (2) the password inputs are detached from the secrets manager
- when the secrets manager is enabled, (1) the secrets existing in the local storage are deleted, and (2) the password input are attached to the secrets manager

jtpio: Cool, sounds good 👍

PR #59, opened 2025-03-18
Avoid building settings schemas when building javascript

brichet:
This remove a leftover script that build the settings schemas when building the jaascript.
 
Follow up #58 which add the settings schemas in the repository.


jtpio (approved):
Thanks!

PR #60, opened 2025-03-18
Do not store passwords to server settings

### LINKED ISSUE #37: Secrets manager

brichet:
This PR adds a plugin providing `ISettingConnector` token, with a `SettingConnector` able to replace the value of some fields from the settings stored on the server (replaced with '***'`).
It is available thanks to a recent change upstream in jupyterlabhttps://github.com/jupyterlab/jupyterlab/pull/17327.

Fixes #37.
Requires https://github.com/jupyterlab-contrib/jupyter-secrets-manager/pull/6 and https://github.com/jupyterlab-contrib/jupyter-secrets-manager/pull/7.

The passwords are not stored anymore in the settings, it is only stored using the `SecretsManager`. 
The previous session is restored by using the settings and the `SecretsManager`.

[record-2025-03-27_18.26.29.webm](https://github.com/user-attachments/assets/cee52c76-f98c-4742-83e5-dc1bb66be472)

### Details

It provides a new `SettingConnector`, which is basically a copy of the one from [jupyterlab](https://github.com/jupyterlab/jupyterlab/blob/main/packages/apputils-extension/src/settingconnector.ts), overriding the save method to remove the password fields. The fields to prevent from storing are populated by the setting panel itself.

When loading a provider from the registry, these fields are replaced by a value fetched from the `SecretsManager`.

It also allows to defer the loading of the provider. When the settings load the first time, it will trigger the loading of the provider in the registry. Since the providers are added to the registry in separate extensions, the settings may load before the availability of the provider. Deferring this load allows to load it as soon as the provider is added to the registry.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
         "@rjsf/utils": "^5.18.4",
         "@rjsf/validator-ajv8": "^5.18.4",
         "json5": "^2.2.3",
-        "jupyter-secrets-manager": "^0.1.1",
+        "jupyter-secrets-manager": "^0.2.0",
         "react": "^18.2.0",
         "react-dom": "^18.2.0"
     },
--- a/package.json
+++ b/package.json
@@ -120,6 +120,8 @@
         "extension": true,
         "outputDir": "jupyterlite_ai/labextension",
         "schemaDir": "schema",
-        "disabledExtensions": ["@jupyterlab/apputils-extension:settings-connector"]
+        "disabledExtensions": [
+            "@jupyterlab/apputils-extension:settings-connector"
+        ]
     }
 }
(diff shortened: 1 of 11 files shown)


brichet:
To test that the provider is automatically enabled with the password when reloading the page, the following patch file uses a local storage connector instead of the default 'in memory' one.

[localStorage.txt](https://github.com/user-attachments/files/19509871/localStorage.txt)

jtpio (approved):
Nice, thanks!

PR #61, opened 2025-03-27
Stop streaming

### LINKED ISSUE #46: Cancel execution / stop streaming responses
### Problem

Users may want to be able to stop the model from generating content, for example if the answer is not satisfying or this there was an error in the prompt.

### Proposed Solution

Implement the approach mentioned in langchain docs to cancel execution: https://js.langchain.com/docs/how_to/cancel_execution/

### Additional context

For the UI element, this would likely require https://github.com/jupyterlab/jupyter-chat/issues/170, unless we can add that button here in an extension directly.

brichet:
Adds a `Stop` button to the input toolbar, that replaces the `Send` button when the bot is writing.

Fixes #46.

Depends on https://github.com/jupyterlab/jupyter-chat/pull/198.

[record-2025-03-25_14.24.10.webm](https://github.com/user-attachments/assets/56213805-fd76-4fff-a464-9ff5d6229547)

Currently it displays an `Aborted` error (new message) in the chat.
We could also handle it by adding something like **...Aborted** at the end of the bot message.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
         "watch:labextension": "jupyter labextension watch ."
     },
     "dependencies": {
-        "@jupyter/chat": "^0.8.1",
+        "@jupyter/chat": "^0.9.0",
         "@jupyterlab/application": "^4.4.0-alpha.0",
         "@jupyterlab/apputils": "^4.5.0-alpha.0",
         "@jupyterlab/completer": "^4.4.0-alpha.0",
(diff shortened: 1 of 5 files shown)


jtpio:
Looking nice!

jtpio (approved):
Nice, thanks!

jtpio:
When testing this with MistralAI as the provider, it looks like messages are still received even after pressing the stop button:


https://github.com/user-attachments/assets/7a66191d-5a82-432c-9077-ac3fb55efc09

Not sure if this is normal for this provider, or if it is an issue with langchain.js. But we can look into this separately.

jtpio:
Starting a new release now.

PR #68, opened 2025-05-05
Fix the API keys in provider when using the secrets manager

### LINKED ISSUE #66: API keys not fetched by the AI provider
## Description

The API keys are not handled properly since #65.

Before the PR, the secrets fields were stored in settings, replacing the value with `"***"`.
The AI provider is catching the configuration fields from the settings to configure the chat and the completer, and replaces the secrets fields by fetching the token from the secrets manager.

If the input is not displayed at all, this field is not stored in the settings, so the provider doesn't know that it has to be fetched.

## Reproduce

Start Jupterlite AI with `Hide secret fields` option checked and try to chat with the bot.

## Expected behavior

The AI provider to be properly configured.

brichet:
Fixes #66

This PR force writing an entry in settings file for each secret, even if the field has been hidden in the settings UI. The value is replaced by `"***"` in the file.
That way, the AI provider registry try to fetch these entries from the secrets manager, and set the provider with the correct secret.

### DIFF
--- a/src/settings/panel.tsx
+++ b/src/settings/panel.tsx
@@ -4,7 +4,6 @@ import {
   ISettingRegistry
 } from '@jupyterlab/settingregistry';
 import { FormComponent, IFormRenderer } from '@jupyterlab/ui-components';
-import { ArrayExt } from '@lumino/algorithm';
 import { JSONExt } from '@lumino/coreutils';
 import { IChangeEvent } from '@rjsf/core';
 import type { FieldProps } from '@rjsf/utils';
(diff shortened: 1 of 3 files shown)


src/settings/panel.tsx
```diff
@@ -137,15 +136,10 @@ export class AiSettings extends React.Component<
     if (!this._secretsManager || !this._useSecretsManager) {
       return;
     }
-    // Attach the password inputs to the secrets manager only if they have changed.
-    const inputs = this._formRef.current?.getElementsByTagName('input') || [];
-    if (ArrayExt.shallowEqual(inputs, this._formInputs)) {
-      return;
-    }
```
brichet, line 144: This comparison has been removed because, some time, when this function is called, some inputs are not yet in the DOM. This leads to a wrong comparison because the number of inputs did not change.
This shouldn't happen, according to the documentation of React. Adding a little timeout fixes it, so I guess there is a race condition (or something related to *RJSF*), but I couldn't fix it properly.

It occurs when updating the form after setting *HideSecretFields* value, which displays again the secrets fields.

It seems that the `await` below is enough for all the element to be in the DOM, but I wonder how robust it is.

brichet:
Merging it to move forward, tested successfully locally.

PR #69, opened 2025-05-12
Add Ollama provider

brichet:
This PR adds the `Ollama` provider to the default ones.

The `Ollama` uses local served model, which adds the possibility for LLM customization and improvements in data security.

**Edit :** It will wait for #70 to be merged, it relies on some of its changes.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,10 +7,13 @@
     "incremental": true,
     "jsx": "react",
     "module": "esnext",
-    "moduleResolution": "bundler",
+    "moduleResolution": "node",
     "noEmitOnError": true,
     "noImplicitAny": true,
     "noUnusedLocals": true,
+    "paths": {
+      "ollama/browser": ["./node_modules/@langchain/ollama/dist"]
+    },
     "preserveWatchOutput": true,
     "resolveJsonModule": true,
     "outDir": "lib",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,7 +7,7 @@
     "incremental": true,
     "jsx": "react",
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
     "noEmitOnError": true,
     "noImplicitAny": true,
     "noUnusedLocals": true,
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,7 +7,7 @@
     "incremental": true,
     "jsx": "react",
     "module": "esnext",
-    "moduleResolution": "bundler",
+    "moduleResolution": "node",
     "noEmitOnError": true,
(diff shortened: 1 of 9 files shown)


tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "incremental": true,
     "jsx": "react",
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
brichet, line 10: This change is required to be able to build the settings for `Ollama`.

It seems that the `"moduleResolution": "node",` is deprecated anyway, and `"bundler"` seems to be a replacement https://www.typescriptlang.org/tsconfig/#moduleResolution.

jtpio: What is the TypeScript issue otherwise?

brichet: The issue is only when building the settings. It uses `ts-json-schema-generator`, which itself requires a `tsconfig` file.
Actually, the simple/safer way is probably to have separate tsconfig.

brichet: > What is the TypeScript issue otherwise?

:smile: I was not expected it after the last commit, but here is the error https://github.com/jupyterlite/ai/actions/runs/15013843059/job/42187366202?pr=69#step:6:73

I think that it is related to `mjs` files that can be imported...

brichet: So it also fails at build time, when importing `@langchain/ollama`, which import `ollama/browser`
```ts
import { Ollama } from "ollama/browser";
```

Here is the typescript error, about importing types from `mts` files :
```
Cannot find module 'ollama/browser' or its corresponding type declarations.
  There are types at '/home/brichet/projects/jupyterlite-ai/node_modules/ollama/dist/browser.d.mts', but this result could not be resolved under your current 'moduleResolution' setting. Consider updating to 'node16', 'nodenext', or 'bundler'.ts(2307)
```

jtpio: I'll have a quick look locally.

jtpio: The relevant issue appears to be https://github.com/ollama/ollama-js/issues/151

jtpio: Maybe we can use the `paths` in `tsconfig.json` for the `ollama/browser` import:

```json
"paths": {
  "ollama/browser": ["./node_modules/@langchain/ollama/dist"]
},
```

jtpio: https://github.com/brichet/jupyterlite-ai/pull/1 appears to be building fine locally.

cc @brichet in case you would like to have a look

brichet:
Thanks @jtpio, I merged your PR and opened #77 to track it.

Should we merge this one ?

jtpio (approved):
Thanks!

PR #70, opened 2025-05-13
Fix completer settings

brichet:
The completer has not worked for several versions, probably since #50.
The reason is that the settings are not properly passed to the completer.

The completers expect an object containing a `settings` props, while the provider registry initialize it with the content of the settings.

https://github.com/jupyterlite/ai/blob/32a67884d09bac4da807b7b1acb86f099a0bff1c/src/base-completer.ts#L40-L42
https://github.com/jupyterlite/ai/blob/32a67884d09bac4da807b7b1acb86f099a0bff1c/src/provider.ts#L187-L189

____

1. This PR fixes it in https://github.com/jupyterlite/ai/commit/3718d4b32cf7cde480d3017c5b532ced92cb5d2c
```ts
        this._completer = new this._currentProvider.completer({
          settings: fullSettings
        });
```

2. It also homogenizes the name of the private variable in the `Completer` to `_completer` instead of `_anthropicProvider`, `chromeProvider`, ...

### DIFF
--- a/src/base-completer.ts
+++ b/src/base-completer.ts
@@ -9,7 +9,7 @@ export interface IBaseCompleter {
   /**
    * The LLM completer.
    */
-  provider: BaseLanguageModel;
+  completer: BaseLanguageModel;
 
   /**
    * The completion prompt.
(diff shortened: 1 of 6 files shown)


jtpio (approved):
Thanks for fixing this!

PR #72, opened 2025-05-13
Align the version of rjsf dependencies

brichet:
This PR is to align `@rjsf/core` to the version of jupyterlab dependency and and to the version of `@rjsf/utils`.


jtpio (approved):
Thanks!

Issue #52, opened 2025-03-06
Warnings in the dev tools console when switching providers


jtpio:
Noticed when testing https://github.com/jupyterlite/ai/pull/48.

Switching the AI provider triggers some warnings to be logged in the dev tools console:

```
[DOM] Found 6 elements with non-unique id #root_smartIndent: (More info: https://goo.gl/9p2vKq) <input type=​"checkbox" id=​"root_smartIndent" name=​"root_smartIndent" aria-describedby=​"root_smartIndent__error root_smartIndent__description root_smartIndent__help" checked>​
```

![Image](https://github.com/user-attachments/assets/047e50f2-7c28-4337-b9de-6fbf9d41d607)

https://github.com/user-attachments/assets/acb9f104-2a82-49d2-be0d-28a5cb3d99c6

brichet:
Thanks for opening this @jtpio, I missed it with Firefox, where these warnings do not seem to be triggered, strangely enough (or maybe I missed a settings in developer tools).

jtpio:
Indeed, the warning is not logged with Firefox (136).

brichet:
There is also 6 elements with these IDs without `jupyterlite-ai`:

<img src=https://github.com/user-attachments/assets/6eab4a31-0bd2-4eef-a58c-f5a0f2abda34 width=500px>

nakul-py:
@jtpio is this issue solved?

PR #73, opened 2025-05-13
Better handling of default values in settings

### LINKED ISSUE #71: Empty a settings text field with default option is not possible
## Description

The settings field with a default value are filled with this value if empty.
That should not prevents user to empty it intentionally.

https://github.com/user-attachments/assets/68481c05-18e4-49d3-b604-738add2d9d6a

## Reproduce

Open the AI settings and try to empty a field with a default value (for example the model field of MistralAI provider).

## Expected behavior

The field to be empty.

brichet:
This PR improve the handling of the default values in the settings panel.

- fix #71 by allowing empty string field when there is a default value
- related to #52 by ensuring that the fields have a unique ID
- display the fields that are modified from the default (using Jupyterlab UI)
- add a button to restore to values to default when relevant (using Jupyterlab UI)

[record-2025-05-13_18.08.23.webm](https://github.com/user-attachments/assets/daf5d9b5-9bc5-40af-9b77-111735949e4d)

### DIFF
--- a/src/settings/panel.tsx
+++ b/src/settings/panel.tsx
@@ -84,7 +84,7 @@ export class AiSettings extends React.Component<
     this._providerSchema = providerSchema as JSONSchema7;
 
     // Check if there is saved values in local storage, otherwise use the settings from
-    // the setting registry (led to default if there are no user settings).
+    // the setting registry (leads to default if there are no user settings).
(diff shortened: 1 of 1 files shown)


jtpio (approved):
Thanks!

PR #75, opened 2025-05-20
Upgrade secrets manager

brichet:
Upgrade `jupyter-secrets-manager` to `>=0.4.0`.

Use the secrets manager new settings to display or not the secrets fields (instead of a jupyterlite-ai setting).
That way there is only one setting to set up all the extensions using the secrets manager.
`jupyter-secrets-manager>=0.4.0` also includes a feature to lock this settings, and therefore prevents third party extension to expose the secrets fields by changing the setting.

This PR also removes the `UseSecretsManager` settings if the extension isn't activated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
         "@rjsf/utils": "^5.18.4",
         "@rjsf/validator-ajv8": "^5.18.4",
         "json5": "^2.2.3",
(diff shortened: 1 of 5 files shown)


trungleduc (approved):

brichet:
Thanks for the review @trungleduc

PR #76, opened 2025-05-21
Update `@langchain/community` to fix ChromeAI

### LINKED ISSUE #74: ChromeAI is broken
## Description

Looks like ChromeAI does not work anymore on the latest Chrome version:

Image

## Reproduce

Tested with https://jupyterlite.github.io/ai/

`window.ai` now seems to be empty:

Image

According to https://developer.chrome.com/docs/ai/debug-gemini-nano, the API seems to have changed:

Image

## Context

Chrome 136.

Langchain.js is still on the previous API: https://github.com/langchain-ai/langchainjs/blob/aeac259a15f68c222108b6090ffb692e0f667fd6/libs/langchain-community/src/experimental/llms/chrome_ai.ts#L165

Not sure if it's worth looking into fixing this in `langchain.js`, if the API breaks again soon (which is likely expected since it's still experimental and behind flags).

jtpio:
Temporary patch for ChromeAI while waiting for https://github.com/langchain-ai/langchainjs/pull/8173.

We can leave https://github.com/jupyterlite/ai/issues/74 open for now to keep track of removing the patch once it's release upstream.

**Edit**: fixed and released upstream.

Fixes #74.


brichet:
Thanks @jtpio, I receive `Uncaught InvalidStateError: The session cannot be created.` when testing it.
Do you know if there is a new setup ?

brichet:
> I receive `Uncaught InvalidStateError: The session cannot be created.` when testing it.

Looks like I ran out of space on my computer, to download the model.

Maybe we should add some notes in the [instruction](https://github.com/jupyterlite/ai/blob/main/src/default-providers/ChromeAI/instructions.ts) about:
- how much space is required (seems to be 22Gb)
- how to debug. This helped me understanding the error https://developer.chrome.com/docs/ai/debug-gemini-nano.
- which minimal version of chrome should be used to have this new API

I also wonder if the current instructions are still relevant, for example I couldn't find the option mentioned in the previous comment :point_up:

brichet:
Otherwise it works well on my side.

jtpio:
A new release of langchain.js is available with the fix: https://github.com/langchain-ai/langchainjs/releases/tag/0.3.27

Will push changes to this PR to simply update to the new packages.

PR #81, opened 2025-05-23
Remove the custom settings connector

brichet:
A custom settings connector has been added in #60, to sanitize the secrets value before saving the settings file on the server.

This PR removes it, in favor of sanitizing the secrets before sending it to the connector. It is done when triggering manually the saving of the settings (when the provider changes or when one of its settings is updated).

The only thing I was worried about (and the reason why we added a custom settings connector) was that the settings are also automatically saved without us triggering it explicitly, and therefore not going through the sanitization. This occurs when updating a setting from the regular setting file.
https://github.com/jupyterlite/ai/blob/82cfd82c30b360b9ec9ecee2debfffa2e71ee6a2/schema/provider-registry.json#L8

But in this case, the settings related to the provider (using our custom component) seems to not be affected.


trungleduc (approved):

brichet:
Thanks for the review @trungleduc

PR #82, opened 2025-05-23
Update Ollama instructions

brichet:
This PR adds some Ollama instructions when serving lite/lab on external server, to fix CORS errors on local Ollama server.


brichet:
The rendering of code block in instructions could be improved by adding at least a background and some bottom padding.



It should be fixed by using the markdown renderer from `jupyter-chat`, which will be isolated with https://github.com/jupyterlab/jupyter-chat/pull/221

jtpio (approved):
Thanks!

PR #84, opened 2025-05-23
Do not expose providers api

brichet:


trungleduc (approved):
Neat!

trungleduc:
And you are tagging the wrong person 😸

trungleduc:
> > to avoid exposing API keys or other secrets that are exposed in langchain model
> 
> That wouldn't prevent a plugin to access `_completer` directly in JavaScript, or with `provider['_completer']`?

you don't have access to the `provider`

brichet:
@trungleduc @jtpio there is now a flag to allow exposing the whole model API when registering the provider.
This flag is used for `WebLLM`, which needs some custom API functions  (there is no sensitive information available in its API)
https://github.com/jupyterlite/ai/blob/d075ec556925566967fc5b65a957338b91f612ab/src/default-providers/index.ts#L157

To prevents being able to modify this flag, the `AIProviders`, `_currentProvider` are not exposed anymore in the the registry.
The `_name` attribute has also been moved in a private namespace, to prevent inconsistency between the name and the models. It could also lead to expose the chat model while not expected.

src/provider.ts
```diff
@@ -180,10 +209,10 @@ export class AIProviderRegistry implements IAIProviderRegistry {
     if (compatibilityCheck !== undefined) {
       const error = await compatibilityCheck();
       if (error !== null) {
-        this._currentProvider = null;
+        // this._currentProvider = null;
```
trungleduc, line 212:
```suggestion
```

jtpio:
Thanks @brichet! Yes exposing less things sounds good.

PR #85, opened 2025-05-24
Improve Mistral completions

jtpio:
Trying out a couple of things to improve the inline suggestions.

- Using `MistralAI` directly (instead of `ChatMistralAI`) gives access to the completion API directly and allows providing more input parameters such as the `prompt`, and `suffix`.
- The use of the throttler seems to be leading to the "character duplication issues". For example typing `df = pd.` would then autocomplete to `df = pd..read_csv()` (with the two dots). Maybe because of the delay or the inline not getting the response for the correct request

https://github.com/user-attachments/assets/9c6a1852-8ea2-49b0-b487-415d69e8c774


jtpio:
> Looks like the `fim` API is not available for all models.

Indeed, I'm mostly testing with `codestral-latest`.

Also since we ship a couple of providers by default, maybe it could make sense to also ship a set of "good defaults", to have the best use of `@jupyterlite/ai` out of the box. For example that could mean choosing some models by default for each of the providers that we know work well. Of course they would still be customizable via the settings.

src/default-providers/MistralAI/completer.ts
```diff
@@ -2,70 +2,54 @@ import {
...
+          suffix
+        },
+        { signal: this._controller.signal },
+        false
+      );
+      const items = response.choices.map(choice => {
+        const content = choice.message.content
+          .replace(CODE_BLOCK_START_REGEX, '')
+          .replace(CODE_BLOCK_END_REGEX, '');
+        return {
```
brichet, line 42: Is this still necessary when using it with the `codestral` model  ?

src/default-providers/MistralAI/completer.ts
```diff
@@ -2,70 +2,54 @@ import {
...
-
     try {
-      return await this._throttler.invoke(messages);
+      const { text, offset: cursorOffset } = request;
+      const prompt = this.systemPrompt.concat(text.slice(0, cursorOffset));
+      const suffix = text.slice(cursorOffset);
+      this._controller.abort();
+      this._controller = new AbortController();
+
+      const response = await this._completer.completionWithRetry(
```
brichet, line 29: Probably for a future PR, but we should handle the status code 400 when the model is not compatible.
It could probably at least be logged in the console log of jupyterlab.

brichet (approved):
Thanks @jtpio, this looks good.

src/default-providers/MistralAI/completer.ts
jtpio: I think we could keep it, just in case?

Also wondering if instead of writing a completer from scratch for all providers, there could be some common logic shared between all the providers, and then each provider would only provide their "fetch" method.

jtpio: > here could be some common logic shared between all the providers

With this I mean sharing utility and post-processing logic, such as cleaning up responses or even caching some results if possible.

PR #92, opened 2025-06-06
Chat panel tweaks

jtpio:
Some minor tweaks for the chat panel:

- Center the welcome message
- restore if `ILayoutRestorer` is available (which may be the case in the next JupyterLite 0.7.0 release)

https://github.com/user-attachments/assets/da5c8b67-31e4-4317-859d-ac36e713461d

### DIFF
--- a/src/chat-handler.ts
+++ b/src/chat-handler.ts
@@ -36,8 +36,6 @@ const AI_AVATAR_BASE64 = btoa(jupyternautLiteIcon.svgstr);
 const AI_AVATAR = `data:image/svg+xml;base64,${AI_AVATAR_BASE64}`;
 
 export const welcomeMessage = (providers: string[]) => `
-<div style="text-align: center; max-width: 350px; margin: 0 auto;">
-
 #### Ask JupyterLite AI
 
 
@@ -47,8 +45,6 @@ the <img src="${AI_AVATAR}" width="16" height="16"> _AI provider_ settings.
 The current providers that are available are _${providers.sort().join('_, _')}_.
 
 To clear the chat, you can use the \`/clear\` command from the chat input.
-
-</div>
 `;
 
 export type ConnectionMessage = {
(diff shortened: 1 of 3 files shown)


src/chat-handler.ts
```diff
@@ -36,6 +36,8 @@ const AI_AVATAR_BASE64 = btoa(jupyternautLiteIcon.svgstr);
 const AI_AVATAR = `data:image/svg+xml;base64,${AI_AVATAR_BASE64}`;
 
 export const welcomeMessage = (providers: string[]) => `
+<div style="text-align: center; max-width: 350px; margin: 0 auto;">
```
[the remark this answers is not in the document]
jtpio: yeah let's do that

style/base.css
```diff
@@ -14,3 +14,9 @@
   color: var(--jp-error-color1);
   font-size: var(--jp-content-font-size1);
 }
+
+.jp-chat-welcome-message {
```
brichet, line 18: Thanks @jtpio

With the last change, the horizontal line is also reduced, I think I prefer a full width but we can fix it later.



To fix it we could:
- move the `hr` tag out of the rendered content (should be done in [jupyter-chat](https://github.com/jupyterlab/jupyter-chat/blob/c9917b74c33e74157ea90867df8915c53d798351/packages/jupyter-chat/src/components/messages/welcome.tsx#L19))
- apply the CSS on `.jp-chat-welcome-message .jp-RenderedMarkdown` (this could be done in this PR or a follow up one).

brichet (commented):
Looks good, thanks @jtpio 

I just have a comment that would need an update in jupyter-chat.

style/base.css
brichet: Yes, I don't have a strong opinion on it, and actually the 3 options (no `hr`, small one or large one) look good.

PR #97, opened 2025-06-16
Add UI tests

### LINKED ISSUE #67: Add UI tests
Currently the repo does not have UI tests.

It would be useful to add a couple of tests to catch potential regressions.

One challenge will be the handling of API keys:

- maybe there could be a way for the UI tests to use an API key set as a GitHub secret?
- otherwise, we could run the tests with a local model, in the browser or with Ollama

brichet:
This PR adds first UI tests, including:
- test the chat panel (button, opening, welcome message)
- error message if there is no AI provider
- AI response if the AI provider is set
- AI code completion if the AI provider is set

Fixes #67


brichet:
The test on code completion seems to work well, but I wonder how stable it will be in the future.
Currently it checks if the ghost text is visible, and also if the request has been send.
If the ghost text fails too often, we can remove the ghost text check.

jtpio:
> The test on code completion seems to work well, but I wonder how stable it will be in the future.
> Currently it checks if the ghost text is visible, and also if the request has been send.
> If the ghost text fails too often, we can remove the ghost text check.

yeah that's also what I was wondering. Otherwise maybe we would need to trigger the completion several time and check that it shows up at least once?

jtpio (approved):
Thanks!

Issue #99, opened 2025-06-17
Gemini models


jtpio:
### Problem

The Gemini family of models is getting more capable lately, for example with Gemini Pro 2.5.

Langchain.js has support for the Google provider: https://js.langchain.com/docs/integrations/platforms/google/

### Proposed Solution

Add Google as one of the providers.

PR #100, opened 2025-06-17
Gemini

jtpio:
Was curious if it would be quick and easy to add Gemini to the list of default providers.

Fixes https://github.com/jupyterlite/ai/issues/99

https://github.com/user-attachments/assets/5a2cc00b-329e-466f-be73-c0c9e4caa46a


package.json
```diff
@@ -65,12 +65,13 @@
...
-        "@langchain/anthropic": "^0.3.9",
-        "@langchain/community": "^0.3.44",
-        "@langchain/core": "^0.3.57",
-        "@langchain/mistralai": "^0.1.1",
-        "@langchain/ollama": "^0.2.0",
-        "@langchain/openai": "^0.4.4",
+        "@langchain/anthropic": "^0.3.22",
+        "@langchain/community": "^0.3.46",
+        "@langchain/core": "^0.3.58",
+        "@langchain/google-genai": "^0.2.12",
```
jtpio, line 71: The latest version of `@langchain/google-genai` seems to depend on newer releases of `@langchain/core`, so bumping the other langchain dependencies here.

tsconfig.json
```diff
@@ -19,6 +19,7 @@
     "resolveJsonModule": true,
     "outDir": "lib",
     "rootDir": "src",
+    "skipLibCheck": true,
```
jtpio, line 22: Need to investigate the following TypeScript build issue:

```
node_modules/zod/v3/index.d.ts:1:15 - error TS2846: A declaration file cannot be imported without 'import type'. Did you mean to import an implementation file '../dist/types/v3/index.js' instead?

1 export * from "../dist/types/v3/index.d.ts";
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/zod/v4/core/index.d.ts:1:15 - error TS2846: A declaration file cannot be imported without 'import type'. Did you mean to import an implementation file '../../dist/types/v4/core/index.js' instead?

1 export * from "../../dist/types/v4/core/index.d.ts";
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

brichet: This seems to be fixed https://github.com/colinhacks/zod/issues/4545
Working fine with `resolutions` pining it to `"^3.25.56"`.

jtpio: Oh cool, thanks!

jtpio: Looks like the latest release may have these changes now: https://github.com/langchain-ai/langchainjs/releases/tag/langchain%3D%3D0.3.29

brichet:
> Looks like there is now an issue when running the `settings:build` script, which may be related to the other `@langchain` package updates.

We should probably drop the settings generator, or at least remove the [test](https://github.com/jupyterlite/ai/blob/c64ad86a0efd34755783c762cd6370a778cc2f85/.github/workflows/build.yml#L93).

I opened https://github.com/jupyterlite/ai/issues/104

brichet:
The failing tests seem to be related to the upgrade of `zod` dependency, after this (huge) PR https://github.com/colinhacks/zod/pull/4074, that has been included in `3.25.X`.
It adds the use of `ThisType` which is not understood by `ts-json-schema-generator`.

I don't know how to fix it, but we have the same issue in https://github.com/jupyterlite/ai/pull/110

jtpio:
> Completions often give a 429 with a lot of requests being made to the Gemini API.

This probably depends on the tier associated with the API key: https://ai.google.dev/gemini-api/docs/rate-limits#free-tier

So for now it seems the basic functionality should work fine. Maybe we can go ahead and continue improving in follow-up PRs.

brichet (approved):
Thanks @jtpio.
Code completion is rather slow on my side, but seems more relevant than with other providers.

PR #101, opened 2025-06-18
Open the settings from the chat panel

jtpio:
Minor improvement to the welcome message, so users can open the settings editor directly from it:

https://github.com/user-attachments/assets/8bfca9d1-08a3-45bf-95f7-fd32cea975ef


brichet:
Nice, I tried to do this but didn't know the `data-commandLinker-command` property.

brichet (approved):
Thanks, looks good.

PR #105, opened 2025-06-20
Allow different providers for the chat and the completer

### LINKED ISSUE #98: Support configuring 2 models, one for the chat and one for the completions
We should think of having 2 models, one for the chat and one for completion.
Looks like the `fim` API is not available for all models.

image

_Originally posted by @brichet in https://github.com/jupyterlite/ai/issues/85#issuecomment-2908939490_

brichet:
This PR allows using different providers for the chat and the completion.
This should improve the user experience, allowing a dedicated provider code completion without degrading chat responses.

The main changes are:
- the provider registry, that has now 2 distinct functions to update the provider settings (whether it is for chat or completer)
- the settings panel, that is now a main component embedding 2 components with all the settings for the providers). When the `UniqueProvider` setting is `true`, only the settings for the chat are displayed, and duplicated for the completer on save.

[output.webm](https://github.com/user-attachments/assets/d7fd181b-85c8-4da4-9b99-e0cd145e1c56)

Fixes #98

### DIFF
--- a/src/settings/panel.tsx
+++ b/src/settings/panel.tsx
@@ -15,7 +15,7 @@ import React from 'react';
 
 import { getSecretId, SECRETS_REPLACEMENT } from '.';
 import baseSettings from './base.json';
-import { IAIProviderRegistry, IDict, ModelUsage, PLUGIN_IDS } from '../tokens';
+import { IAIProviderRegistry, IDict, ModelRole, PLUGIN_IDS } from '../tokens';
 
 const MD_MIME_TYPE = 'text/markdown';
 const INSTRUCTION_CLASS = 'jp-AISettingsInstructions';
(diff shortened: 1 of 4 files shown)


src/settings/panel.tsx
```diff
@@ -510,6 +689,7 @@ export class AiSettings extends React.Component<
     );
   }
 
+  private _usage: modelUsage;
```
brichet, line 692: Perhaps we could improve the name of the choice variable between `chat` and `completion`.
Currently, the type is called `modelUsage` and the variable `_usage` (in several places), I haven't found anything more relevant.

schema/provider-registry.json
```diff
@@ -11,10 +11,16 @@
       "description": "Whether to use or not the secrets manager. If not, secrets will be stored in the browser (local storage)",
       "default": true
     },
-    "AIprovider": {
+    "UniqueProvider": {
+      "type": "boolean",
+      "title": "One provider for chat and completer",
```
jtpio, line 16: Maybe something like the following could also work?

```suggestion
      "title": "Use the same provider for chat and completer",
```

src/tokens.ts
```diff
@@ -15,6 +15,8 @@ export const PLUGIN_IDS = {
   systemPrompts: '@jupyterlite/ai:system-prompts'
 };
 
+export type modelUsage = 'chat' | 'completer';
```
jtpio, line 18: Maybe this type can also start with an uppercase letter, for consistency with the JupyterLab code base?

```suggestion
export type ModelUsage = 'chat' | 'completer';
```

src/settings/panel.tsx
jtpio: I meant having both

```ts
get currentChatName(): string 
```

and

```ts
get currentCompleterName(): string
```

as an alternative to:

```ts
currentName(usage: ModelUsage): string
```

If that can help avoid having to expose a `ModelUsage` type.

brichet: Not sure to get it.
AFAIU, what you suggest would probably be enough for the provider registry, but we would have to duplicate also the functions that save settings in the registry and in local storage (https://github.com/jupyterlite/ai/pull/105/files#diff-c163aac3bfa8d91d61fc51c382f75f20f4aa46ef650e860db9d52b8a65a3670aR52).
Do you think that it would be a better implementation ?

brichet: I updated it (and rebased from main)

PR #110, opened 2025-06-26
Use langgraph (agent) to call tools

brichet:
This PR adds [`langgraph`](https://langchain-ai.github.io/langgraphjs/), which allows to call tools that can interact with Jupyterlab API.

As an example, the PR includes a default tools, that only shows a Jupyterlab modal with the response from the model.

UPDATED RECORDING

[output.webm](https://github.com/user-attachments/assets/18dc39b9-5a1d-4920-9675-0a993de7626d)

~~[record-2025-06-26_17.23.44.webm](https://github.com/user-attachments/assets/8332da68-b34f-47ee-8e88-5989a1d78dbe)~~

## Current state
- add a setting to enable the use of tool (only a test tool available as of now)
- create an agent (LLM + tool) using `langgraph`, and call it instead of only the LLM
- the LLM decides by itself if it should call a tool or not. The PR updates the system prompt to require calling it, for now.
- add a message for each message received from the LLM (user is `Jupyternaut`) and from the tool (user is the tool name)

## To do, To discuss
- [x] add a registry (with a token) to enable additional tools from extensions
- [x] @jtpio suggested to switch from LLM to agent from the chat panel itself (instead of the settings). If we do that, we should probably be able to select the tool(s) to use from the chat panel too.
- [x] currently there is an additional instruction in the system prompt to request the use of the tool. We should explore if there is an option when binding the tool to the model, if we want to force the use of it. **EDIT:** *this additional instruction has been removed, the model should have now enough information about the capabilities of the tool in the description to decide to use it or not*
- [x] we should probably pass only a subset of the agent for security reason (as we did for chat model and completer model in https://github.com/jupyterlite/ai/pull/84)
- [x] fix the message history (or maybe the prompt), to avoid redoing the tool actions on each new message.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
         "watch:labextension": "jupyter labextension watch ."
     },
     "dependencies": {
-        "@jupyter/chat": "^0.14.0",
+        "@jupyter/chat": "^0.17.0",
         "@jupyterlab/application": "^4.4.0",
         "@jupyterlab/apputils": "^4.5.0",
         "@jupyterlab/completer": "^4.4.0",
--- a/schema/chat.json
+++ b/schema/chat.json
@@ -22,6 +22,12 @@
       "title": "AI persona name",
       "description": "The name of the AI persona",
       "default": "Jupyternaut"
+    },
+    "enableClearChatButton": {
+      "type": "boolean",
+      "title": "Enable Clear Chat Button",
+      "description": "Whether to enable the button to clear the chat history.",
+      "default": false
     }
   },
   "additionalProperties": false
--- a/src/index.ts
+++ b/src/index.ts
@@ -83,16 +83,25 @@ const chatPlugin: JupyterFrontEndPlugin<void> = {
     let sendWithShiftEnter = false;
     let enableCodeToolbar = true;
     let personaName = 'AI';
+    let enableClearChatButton = false;
 
     function loadSetting(setting: ISettingRegistry.ISettings): void {
       sendWithShiftEnter = setting.get('sendWithShiftEnter')
         .composite as boolean;
       enableCodeToolbar = setting.get('enableCodeToolbar').composite as boolean;
-      personaName = setting.get('personaName').composite as string;
+      personaName = setting.get('personaName')?.composite as string;
+      enableClearChatButton = setting.get('enableClearChatButton')
+        .composite as boolean;
 
       // set the properties
       chatHandler.config = { sendWithShiftEnter, enableCodeToolbar };
       chatHandler.personaName = personaName;
+
+      if (enableClearChatButton && chatHandler.messages.length > 0) {
+        inputToolbarRegistry.show('clear');
+      } else {
+        inputToolbarRegistry.hide('clear');
+      }
     }
 
     Promise.all([app.restored, settingsRegistry?.load(chatPlugin.id)])
@@ -142,6 +151,10 @@ const chatPlugin: JupyterFrontEndPlugin<void> = {
     });
 
     chatHandler.messagesUpdated.connect(() => {
+      if (!enableClearChatButton) {
+        inputToolbarRegistry.hide('clear');
+        return;
+      }
       if (chatHandler.messages.length > 0) {
         inputToolbarRegistry.show('clear');
       } else {
--- a/yarn.lock
+++ b/yarn.lock
@@ -843,9 +843,9 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@jupyter/chat@npm:^0.14.0":
-  version: 0.14.0
-  resolution: "@jupyter/chat@npm:0.14.0"
+"@jupyter/chat@npm:^0.17.0":
+  version: 0.17.0
+  resolution: "@jupyter/chat@npm:0.17.0"
   dependencies:
     "@emotion/react": ^11.10.5
     "@emotion/styled": ^11.10.5
@@ -870,7 +870,7 @@ __metadata:
     clsx: ^2.1.0
     react: ^18.2.0
     react-dom: ^18.2.0
-  checksum: a3a4e4156cac707a70f7e9d77ef0bedf4000fcd052a7dd3d6a13f210e160d0ebfc5b03837ba12efe134aa1cde1e76755e01a48887e9fc4e3db94a80bdc760d22
+  checksum: 75d906e59c28ae34132011583769bab2ce7e2d727888c7e644f2db0ddda60d9bf46a9c16bac67e55d06ad82319078ec0f6eed5468534bde5fd3c15a279e8be45
   languageName: node
   linkType: hard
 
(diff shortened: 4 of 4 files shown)


brichet:
I open this PR to move forward, and to allow working on useful tools in a separate PR.
When we have some tools that are useful, we'll probably want to remove the current test tool.

I also wonder if should allow disabling some tools from the settings (in future PR maybe), to avoid displaying them in the list, if a lot are provided by extensions.

src/tools/create-notebook.ts
```diff
@@ -0,0 +1,37 @@
...
+      schema: z.object({
+        command: z.string().describe('The Jupyterlab command id to execute'),
+        args: z
+          .object({})
+          .passthrough()
+          .describe('The argument for the command')
+      })
+    }
+  );
+};
```
jtpio, line 37: Maybe we could pass the `kernelName` with something like the following:


```suggestion
export const createNotebook = (
  commands: CommandRegistry
): StructuredToolInterface => {
  return tool(
    async ({ command, args, kernelName }) => {
      let result: any = 'No command called';
      if (command === 'notebook:create-new') {
        // Create args object with kernelName if provided
        const commandArgs: ReadonlyPartialJSONObject = {
          ...args,
          ...(kernelName && { kernelName })
        };

        result = await commands.execute(command, commandArgs);
      }
      const output = `
The test tool has been called, with the following query: "${command}"
The args for the commands where ${JSON.stringify(args)}
${kernelName ? `The kernel name specified: "${kernelName}"` : ''}
The result of the command (if called) is "${result}"
`;
      return output;
    },
    {
      name: 'createNotebook',
      description: 'Run jupyterlab command to create a notebook',
      schema: z.object({
        command: z.string().describe('The Jupyterlab command id to execute'),
        args: z
          .object({})
          .passthrough()
          .describe('The argument for the command'),
        kernelName: z
          .string()
          .optional()
          .default('python3')
          .describe(
            'The name of the kernel to use for the notebook. If not specified, will use the default kernel.'
          )
      })
    }
  );
};
```

jtpio:
The extension seems to be building fine locally, although it leaves some kind of error / warnings in the logs:

```
WARNING in ./node_modules/@langchain/langgraph/dist/pregel/validate.js
Invalid dependencies have been reported by plugins or loaders for this module. All reported dependencies need to be absolute paths.
Invalid dependencies may lead to broken watching and caching.
As best effort we try to convert all invalid values to absolute paths and converting globs into context dependencies, but this is deprecated behavior.
Loaders: Pass absolute paths to this.addDependency (existing files), this.addMissingDependency (not existing files), and this.addContextDependency (directories).
Plugins: Pass absolute paths to fileDependencies (existing files), missingDependencies (not existing files), and contextDependencies (directories).
Globs: They are not supported. Pass absolute path to the directory as context dependencies.
The following invalid values have been reported:
 * "../../src/pregel/validate.ts"
 @ ./node_modules/@langchain/langgraph/dist/pregel/index.js 20:0-60 439:8-21 1149:12-24 1156:12-24
 @ ./node_modules/@langchain/langgraph/dist/graph/graph.js 7:0-53 329:35-41 368:32-51
 @ ./node_modules/@langchain/langgraph/dist/graph/state.js 5:0-59 82:32-37 470:40-53 777:15-21
 @ ./node_modules/@langchain/langgraph/dist/prebuilt/agent_executor.js 2:0-47 39:25-35
 @ ./node_modules/@langchain/langgraph/dist/prebuilt/index.js 1:0-59 1:0-59
 @ ./node_modules/@langchain/langgraph/prebuilt.js 1:0-40 1:0-40
 @ ./lib/provider.js 1:0-65 329:29-45
 @ ./lib/index.js 14:0-48 138:37-55
 @ container entry ./extension[0] ./index[0]
```

src/provider.ts
```diff
@@ -374,6 +390,29 @@ export class AIProviderRegistry implements IAIProviderRegistry {
...
+   * Build an agent with a given tool.
+   */
+  buildAgent(tools: Tool[]) {
+    if (tools.length) {
+      const chatModel = Private.getChatModel();
+      if (chatModel === null) {
+        Private.setAgent(null);
+        return;
+      }
+      chatModel.bindTools?.(tools);
```
jtpio, line 403: Maybe we'll need to show a message somewhere that the selected provider does not support tool calling, or disable the functionality automatically if it can be detected.

For example when trying to use tools with ChromeAI:

image

brichet: I updated the PR.

- the `allowTools` setting is now in the provider registry settings instead of chat settings, for consistency (the agent is built there).
- the set tools button is now disabled if the provider cannot handle tools, or if the provider is not set (e.g. no APIkey).

Unfortunately, it seems that it does not work at the model level. The tools are still available for example with Claude 2.1. When sending a message, we receive an error message (`'claude-2.1' does not support tool use via API fields.`)

brichet:
> The extension seems to be building fine locally, although it leaves some kind of error / warnings in the logs:

Yes, I saw it, it seems to be a issue with the packaging of `@langchain/langgraph`.

jtpio:
> Maybe we can handle these warnings with a webpack setting...

Yeah providing a custom webpack config to ignore these warnings seem to be do the trick for now: https://github.com/brichet/jupyterlite-ai/pull/2

jtpio:
Thanks @brichet!

Just tried quickly again with the latest state and it looks like a good start 👍 

Spotted a few small things. The tool greyed out icon looks like the feature is not enabled and would not allow selecting tools, although it does. Maybe it should have the same color as the send button?

https://github.com/user-attachments/assets/5bd6346e-29df-4fd1-8478-762e729dbb9b

Ideally we should be able to see the list of arguments instead of `[object Object]`:

image

And there is likely something to do with the message history too, as the model seems to be taking into account previous requests:

image

But I guess it would be ok to fix that separately too.

brichet:
Trying to rebase to resolve the conflicts, I can't build anymore:
```bash
node_modules/@langchain/langgraph/dist/graph/messages_annotation.d.ts:82:15 - error TS2589: Type instantiation is excessively deep and possibly infinite.

82     messages: import("./zod/meta.js").ReducedZodChannel<z.ZodType<BaseMessage[], z.ZodTypeDef, BaseMessage[]>, import("@langchain/core/utils/types").InteropZodType<Messages>>;
                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/tools/create-notebook.ts:9:3 - error TS2589: Type instantiation is excessively deep and possibly infinite.

9   return tool(
    ~~~~~~

src/tools/create-notebook.ts:9:10 - error TS2589: Type instantiation is excessively deep and possibly infinite.

  9   return tool(
             ~~~~~
 10     async ({ command, args }) => {
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
... 
 35     }
    ~~~~~
 36   );
    ~~~


Found 3 errors in 2 files.

Errors  Files
     1  node_modules/@langchain/langgraph/dist/graph/messages_annotation.d.ts:82
     2  src/tools/create-notebook.ts:9
```

jtpio:
> I don't know if this possible, but we should think of a way to prevent such interactions during the LLM response (except for stopping the LLM).

UX wise, it would still be nice to let users interact with the UI while the agent is performing actions. 

So maybe this would mean having some wrapper commands that would some data as input / args, such as the notebook to perform actions on, the cell to edit and so on?

PR #113, opened 2025-06-27
Remove provider settings check/generation

### LINKED ISSUE #104: Drop settings checker in tests
We should probably drop the settings generator, or at least remove the [test](https://github.com/jupyterlite/ai/blob/c64ad86a0efd34755783c762cd6370a778cc2f85/.github/workflows/build.yml#L93), since it brings several issues:
- https://github.com/jupyterlite/ai/pull/100#issuecomment-2984134456
- https://github.com/jupyterlite/ai/issues/64

This generator depends on [ts-json-schema-generator](https://github.com/vega/ts-json-schema-generator), which doesn't seems to handle correctly nested types and functions in our case.

We could still keep the script to generate manually a first version of the settings for a specific provider, like
```bash
jlpm settings:build OpenAI
```

brichet:
This PR remove the automatic use of provider settings checker/generation.

Currently the script that perform this check/generation is still there.
We should probably remove it as well if we don't expect to fix it (see below for current failures).

> The failing tests seem to be related to the upgrade of `zod` dependency, after this (huge) PR https://github.com/colinhacks/zod/pull/4074, that has been included in `3.25.X`.
> It adds the use of `ThisType` which is not understood by `ts-json-schema-generator`.
> 
> I don't know how to fix it, but we have the same issue in https://github.com/jupyterlite/ai/pull/110

_Originally posted by @brichet in https://github.com/jupyterlite/ai/issues/100#issuecomment-3012721937_

### Related issues

Fixes #104
Related to test failures in #100 and #110.


jtpio (approved):
Makes sense, thanks!

PR #131, opened 2025-07-23
Adding New chat button

### LINKED ISSUE #111: Ability to create new chats and move them to the main area
### Problem

Currently there is only one chat panel, and the only way to get a clean state is to send the `/clear` message:

Image

### Proposed Solution

Something like in jupyter chat:

Image

nakul-py:
Adding `+ Chat` button into the Jupyterlite Ai chat with the help of creating new header for button.

https://github.com/user-attachments/assets/b69f4d1c-30e5-461f-9248-7e7b2f728fe9

- Closes #111

### DIFF
--- a/src/components/chat-header.tsx
+++ b/src/components/chat-header.tsx
@@ -1,0 +1,45 @@
+import React, { useEffect, useRef } from 'react';
+import { ReactWidget } from '@jupyterlab/apputils';
+import { buildChatSidebar, IChatModel } from '@jupyter/chat';
+import { NewChatButton } from './new-chat-button';
+import { jupyternautLiteIcon } from '../icons';
+
+export function buildSidebarWithHeader(
+  options: Parameters<typeof buildChatSidebar>[0],
+  newChat: () => void,
+  model: IChatModel
+): ReactWidget {
+  const sidebar = buildChatSidebar(options);
+
+  const ChatSidebarWithHeader = () => {
+    const containerRef = useRef<HTMLDivElement>(null);
+
+    useEffect(() => {
+      if (containerRef.current) {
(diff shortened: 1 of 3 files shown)


src/components/chat-header.tsx
```diff
@@ -0,0 +1,45 @@
...
+        >
+          <NewChatButton newChat={newChat} model={model.input} />
+        </div>
+        <div style={{ flex: 1, overflow: 'auto' }} ref={containerRef}></div>
+      </div>
+    );
+  };
+
+  const widget = ReactWidget.create(<ChatSidebarWithHeader />);
+  widget.title.icon = jupyternautLiteIcon;
```
nakul-py, line 43: I use this Icon here because creating chatsidebar with header changes the Jupyterlite icon. So, I use this already created JupyterLite AI chat icon after successfully creating a chatsidebar with a header. 

@jtpio and @brichet you have any other way to get back the old icon after rendering chatsidebar? Please share the idea.

brichet:
Thanks @nakul-py for working into this.
If I understand correctly this PR, there would be one chat in the side panel, and the other ones would be created in the main area ? 

I wonder if we could find a way to reuse the multi-chats panel from `jupyterlab-chat`.
This panel could probably be moved to the `@jupyter/chat` package and be reused in `jupyterlab-chat` and `@jupyterlite/ai`.

nakul-py:
@brichet will you reference the main-chats panel from https://github.com/jupyterlab/jupyter-chat and @jtpio your suggestions will helps too.

brichet:
I opened https://github.com/jupyterlab/jupyter-chat/issues/261 to track it in jupyter-chat.

PR #132, opened 2025-08-04
Adding Clear button

### LINKED ISSUE #130: Clear Button
### Problem
 Users always need to type `/clear` command to clear the chat.

### Proposed Solution

Want to add a clear chat button. Here is how it works...

https://github.com/user-attachments/assets/89d66bb3-9947-4ff4-aa73-c6f2c1b4fc19

### Additional context

@brichet and @jtpio, you can suggest a better position for the clear button and other nitpicks.

nakul-py:
https://github.com/user-attachments/assets/377fb33a-5e19-489d-8322-e06b02d4f01f

Closes #130

### DIFF
--- a/src/chat-handler.ts
+++ b/src/chat-handler.ts
@@ -62,8 +62,8 @@ export class ChatHandler extends AbstractChatModel {
     });
   }
 
-  clearChat(): void {
-    this.clearMessages();
+  clearMessages(): void {
+    super.clearMessages();
     this._history.messages = [];
   }
 
--- a/src/chat-handler.ts
+++ b/src/chat-handler.ts
@@ -62,6 +62,11 @@ export class ChatHandler extends AbstractChatModel {
     });
   }
 
+  clearChat(): void {
+    this.messagesDeleted(0, this.messages.length);
+    this._history.messages = [];
+  }
+
   get provider(): AIChatModel | null {
     return this._providerRegistry.currentChatModel;
   }
--- a/src/chat-handler.ts
+++ b/src/chat-handler.ts
@@ -63,7 +63,7 @@ export class ChatHandler extends AbstractChatModel {
   }
 
   clearChat(): void {
-    this.messagesDeleted(0, this.messages.length);
+    this.clearMessages();
     this._history.messages = [];
   }
 
--- a/src/components/clear-button.tsx
+++ b/src/components/clear-button.tsx
@@ -2,8 +2,7 @@
  * Copyright (c) Jupyter Development Team.
  * Distributed under the terms of the Modified BSD License.
  */
-
-import BackspaceIcon from '@mui/icons-material/Backspace';
+import CleaningServicesRoundedIcon from '@mui/icons-material/CleaningServicesRounded';
 import React from 'react';
 
 import { InputToolbarRegistry, TooltippedButton } from '@jupyter/chat';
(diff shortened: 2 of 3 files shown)


src/chat-handler.ts
```diff
@@ -62,6 +62,11 @@ export class ChatHandler extends AbstractChatModel {
     });
   }
 
+  clearChat(): void {
```
brichet, line 65: The `AbstractChatModel` provides a `clearMessages()` method, which should delete all the message and trigger a signal. We should probably keep its name and call it here.

src/index.ts
```diff
@@ -127,6 +136,14 @@ const chatPlugin: JupyterFrontEndPlugin<void> = {
       }
     });
 
+    chatHandler.messageAdded = (message: IChatMessage): void => {
```
brichet, line 139: We should make use of the [`messageUpdated`](https://github.com/jupyterlab/jupyter-chat/blob/98898f104681b85808a477af2b91d10a3a49745c/packages/jupyter-chat/src/model.ts#L86) signal to toggle the visibility of the button.

src/components/clear-button.tsx
```diff
@@ -0,0 +1,56 @@
...
+    <TooltippedButton
+      onClick={props.clearChat}
+      tooltip={tooltip}
+      buttonProps={{
+        size: 'small',
+        variant: 'contained',
+        title: tooltip
+      }}
+    >
+      <BackspaceIcon />
```
brichet, line 38: Not sure about that icon. I wonder if there some more explicit icon.

brichet (changes requested):
Thanks @nakul-py for working on this, it works well.

src/components/clear-button.tsx
nakul-py: Currently using this [formatclear icon](https://mui.com/material-ui/material-icons/?query=formatcl&selected=FormatClear) and you can suggest better icon from [Icons](https://mui.com/material-ui/material-icons/#search-material-icons)

brichet: Or maybe [this one](https://github.com/jupyterlab/jupyterlab/blob/main/packages/ui-components/style/icons/toolbar/cleaning.svg) from jupyterlab, which seems to be designed for cleaning.

src/index.ts
nakul-py: done

src/components/clear-button.tsx
```diff
@@ -0,0 +1,55 @@
...
+ */
+export function clearItem(
+  clearChat: () => void
+): InputToolbarRegistry.IToolbarItem {
+  return {
+    element: (props: InputToolbarRegistry.IToolbarItemProps) => {
+      const clearProps: IClearButtonProps = { ...props, clearChat };
+      return <ClearButton {...clearProps} />;
+    },
+    position: 60
```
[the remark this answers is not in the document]
nakul-py: done

nakul-py:
> Not sure whether we should keep this command or not, let's do that in an other PR.

will you open a new issue on this and mention @jtpio for their opinion too.

brichet (approved):
Thanks @nakul-py, this looks good to me.
Just a nitpick comment below, let me know what you think.

brichet:
Thanks @nakul-py

PR #138, opened 2025-08-08
Update langchain dependencies

jtpio:
Mostly to stay up-to-date with recent releases.


jtpio:
The build passes, but I haven't checked all providers still work as expected.

jtpio:
Did some basic testing with some providers and did not notice anything breaking with this update.

PR #152, opened 2025-09-16
Refactoring

brichet:
Some refactoring on the code to:

- split the main plugin to provide a settings-model and a tool-registry plugins
- use the provider ID instead of the name to retrieve the API key in settings model
- use the chat context in the input toolbar buttons (required if we move to multichat panel)


nakul-py:
Thanks @brichet for the explanation. :)

PR #157, opened 2025-09-17
Fix parameters for `discover_commands`

jtpio:
For some reasons this was triggering an error on tool call when using Claude Sonnet 4.

For now we can put a placeholder to later be able to query commands.


brichet (approved):
Thanks @jtpio, I confirm that it fixes the bug with claude sonnet

Issue #146, opened 2025-09-12
Switch to ai-sdk and openai-agent


brichet:
Following https://github.com/jupyterlite/ai/pull/145, to track the missing features.

- [x] Secret manager
- [x] UI tests
- [ ] advanced provider settings
- [x] missing default providers
  - [x] openAI: https://github.com/jupyterlite/ai/pull/147
  - [ ] ...
- [ ] refactoring / cleaning
  - [ ] settings panel
  - [ ] ...

brichet:
> * advanced provider settings

The `Model Parameters` tab should probably be in the `Edit Provider` dialog, to be able to fine tune each provider independently.

jtpio:
Noting that the AI SDK recently introduced their own `Agent` wrapper: https://ai-sdk.dev/docs/agents/overview

Which will likely be available in the upcoming 5.1 release.

Some other agent related hook such as approval flow also seem to have been added recently: https://github.com/vercel/ai/releases/tag/%40ai-sdk%2Fprovider-utils%403.1.0-beta.8

So maybe it could be worth considering using this `Agent` abstraction when it's ready, which would help drop one dependency (in this case `openai-agent-js`).

PR #159, opened 2025-09-18
Add the secrets manager

brichet:
This PR reintroduces the use of the secrets manager in the `next` branch.

Fixes partially #146 

[output.webm](https://github.com/user-attachments/assets/285611a6-3937-44ed-b17c-b1120730aa34)

### DIFF
--- a/schema/settings-model.json
+++ b/schema/settings-model.json
@@ -4,6 +4,12 @@
   "description": "Configuration for JupyterLite AI extension providers, models, and behavior",
   "type": "object",
   "properties": {
+    "useSecretsManager": {
+      "type": "boolean",
+      "title": "Use secrets manager",
+      "description": "Whether to use or not the secrets manager. If not, secrets will be stored in the browser (local storage)",
+      "default": true
+    },
     "providers": {
       "title": "AI Providers",
       "description": "List of configured AI providers",
--- a/schema/settings-model.json
+++ b/schema/settings-model.json
@@ -7,7 +7,7 @@
     "useSecretsManager": {
       "type": "boolean",
       "title": "Use secrets manager",
-      "description": "Whether to use or not the secrets manager. If not, secrets will be stored in the browser (local storage)",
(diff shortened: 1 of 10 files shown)


schema/settings-model.json
```diff
@@ -4,6 +4,12 @@
   "description": "Configuration for JupyterLite AI extension providers, models, and behavior",
   "type": "object",
   "properties": {
+    "useSecretsManager": {
+      "type": "boolean",
+      "title": "Use secrets manager",
+      "description": "Whether to use or not the secrets manager. If not, secrets will be stored in the browser (local storage)",
```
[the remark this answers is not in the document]
brichet: Right, I can update the description.

schema/settings-model.json
```diff
@@ -4,6 +4,12 @@
   "description": "Configuration for JupyterLite AI extension providers, models, and behavior",
   "type": "object",
   "properties": {
+    "useSecretsManager": {
+      "type": "boolean",
+      "title": "Use secrets manager",
+      "description": "Whether to use or not the secrets manager. If not, secrets will be stored in plain text in settings",
+      "default": true
```
jtpio, line 11: Wondering if this should default to `false`? 

Looks like by default the secret manager does not use a connector, so that means API keys are lost after refreshing the page and need to be entered again?

brichet: > Probably we will need to document the usage of the Secret Manager a little bit more.
:+1: 

> Looks like by default the secret manager does not use a connector, so that means API keys are lost after refreshing the page and need to be entered again?

The default is using an "in memory" connector, so indeed it is lost after reloading the page.
But I wonder if this behavior isn't better than storing the API key in plain text by default.

jtpio: Right, maybe it makes more sense like this.

And it can always be disabled via the UI or the settings.

jtpio (approved):
Thanks!

PR #162, opened 2025-09-22
Call the diff command directly when setting the cell content

jtpio:
This makes the diff show up instantly, instead of having to wait for the LLM to get the list of commands and then execute the command to show the diff:

https://github.com/user-attachments/assets/8c8b4a7c-ff06-4026-b215-52f44bebd9e5

### DIFF
--- a/src/agent.ts
+++ b/src/agent.ts
@@ -760,19 +760,16 @@ Guidelines:
 - End with a brief summary of accomplishments
 - Use natural, conversational tone throughout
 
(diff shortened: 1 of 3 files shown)


brichet:
Some ideas for follow up PR::
- add settings to disable the suggestion mode
- require approval to apply the change (instead of applying it and allow to undo)
- show new cell as a diff too
  
  [output.webm](https://github.com/user-attachments/assets/a999d9ad-42b2-49e3-b6b3-289a07ca88e3)

Issue #144, opened 2025-09-08
Support using a server side proxy to forward requests to AI providers


jtpio:
### Problem

Currently API keys are managed in the browser (or via https://github.com/jupyterlab-contrib/jupyter-secrets-manager), and the requests are also made from the browser.

This is convenient and makes it really easy to use an existing provider without running anything extra locally.

However, this can lead to some potential security issues since the API keys may be stored in the browser local storage.

### Proposed Solution

There should be a way to allow proxying the requests to a component or proxy running on a server. That component and/or proxy will be responsible for forwarding the request to the corresponding AI provider with the proper API keys.

When using this extension in stock JupyterLab, this could take the form of a simple Jupyter Server extension.

With JupyterLite, this could be a dedicated Node.js or Python application that could easily be hosted on serverless platforms like Vercel, Netlify, or Cloudflare Workers.

### Additional context

In any case, this should be optional and easily configurable.

PR #171, opened 2025-09-30
Custom base URL and generic provider (OpenAI compatible)

jtpio:
Fixes https://github.com/jupyterlite/ai/issues/144

- Allow providing a custom baser URL when configuring providers
- Add a new generic provider (OpenAI compatible)

image

For example this allows using the LiteLLM simple proxy (https://docs.litellm.ai/docs/simple_proxy) with the generic OpenAI compatible provider: 

https://github.com/user-attachments/assets/2c0433e0-f806-48e3-b4cf-43861a1c5a59

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,52 @@ Then restart Chrome for these changes to take effect.
 > [!NOTE]
 > For more information about Chrome Built-in AI: https://developer.chrome.com/docs/ai/get-started
 
+## Using LiteLLM Proxy
+
+[LiteLLM Proxy](https://docs.litellm.ai/docs/simple_proxy) is an OpenAI-compatible proxy server that allows you to call 100+ LLMs through a unified interface.
+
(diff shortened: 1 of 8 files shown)


src/widgets/ai-settings.tsx
```diff
@@ -621,26 +621,36 @@ const AISettingsComponent: React.FC<IAISettingsComponentProps> = ({
...
+                      return (
+                        <ListItem key={provider.id} divider>
+                          <ListItemText
+                            primary={provider.name}
+                            secondary={
+                              <Typography
+                                variant="body2"
+                                color="text.secondary"
+                              >
+                                {provider.provider} • {provider.model}
```
brichet, line 637:
```suggestion
                                {provider.provider} • {provider.model}
```

Is this used somewhere ?

brichet (approved):
Thanks @jtpio.
I haven't tested it yet but the code looks good.

I wonder if we should not add some general description in the provider settings dialog at some point. This could help being more descriptive about the use of this provider, for example.

README.md
```diff
@@ -90,6 +90,52 @@ Then restart Chrome for these changes to take effect.
 > [!NOTE]
 > For more information about Chrome Built-in AI: https://developer.chrome.com/docs/ai/get-started
 
+## Using LiteLLM Proxy
```
jtpio, line 93: Added a bit of documentation to cover the use case mentioned in https://github.com/jupyterlite/ai/issues/144

brichet:
Thanks @jtpio

PR #175, opened 2025-10-06
Token usage accumulation

### LINKED ISSUE #173: Toekn usage: the values are reset for each request
Currently the values of the token usage are reset for each request.

We should instead add the values to the previous ones, it may interesting to know the token usage for the whole chat request.

brichet:
Fixes #173


jtpio (approved):
Thanks!

PR #176, opened 2025-10-06
Switch to ai-sdk and openai-agent

brichet:
The `next` branch should now be ready, and improves a lot the user experience comparing to the current stable version.
Event if some of the items from #146 are not fulfilled, we can probably merge it in `main` and make improvements from there.

There is also a [0.8.x](https://github.com/jupyterlite/ai/tree/0.8.x) that should be used to update the current stable version.


jtpio:
Looks like we could go with this merge into `main` before https://github.com/jupyterlite/ai/pull/169? For https://github.com/jupyterlite/ai/pull/169 it would be changing the target branch to `main` afterwards.

jtpio (approved):
Thanks!

Issue #64, opened 2025-04-29
Allow custom API url endpoint for OpenAI compatible API service


ryuzio:
### Problem

Besides given AI service providers, there are also 3rd party providers for LLM inference (such as OpenRouter). Some people are self-hosting their LLMs as well (Ollama for example). It would be handy if such service providers could be available in JupyterLite AI.

### Proposed Solution

Providing a config item for custom API endpoint to use. It could be part of OpenAI config, or as a separate custom provider.

ryuzio:
I checked the source code, and saw `@langchain/openai` is used. According to the doc of the package, it seems that it already provides capability to [instantiate the client with custom urls](https://js.langchain.com/docs/integrations/chat/openai/#custom-urls). So I guess what needs to be done is to expose `baseURL` in the settings of JupyterLite AI.

I'm not a JS developer so I'm not 100% sure. Some insights from more experienced developers to see if it is something that could be introduced in future versions of JupyterLite AI would be appreciated.

brichet:
Thanks @ryuzio for looking into this.

Currently the settings schema for each provider are automatically generated from the provider options interface.
This is done with [this script](https://github.com/jupyterlite/ai/blob/main/scripts/settings-checker.js), using the https://github.com/vega/ts-json-schema-generator library.

The benefit of this automation is to ensure that the schema is up to date with the current version of the provider, and to avoid having to set manually each field.
The drawback is that some settings are not available because it does not ft with a JSON schema (function, objects...).

For OpenAI, we had to remove the configuration prop that was breaking this schema generation
https://github.com/jupyterlite/ai/blob/830b2c33f15e7ca406e429cbb4398f89bcd8d253/scripts/settings-checker.js#L61
I think that this prop contains the `baseURL` you are mentioning.

We should probably find a way to be more specific on the prop to exclude (the `baseURL` could be included, it is only a string), or to handle all the props.

pmkrawczyk:
In addition to allowing OpenAI-compatible LLM providers, this would open a possibility for using custom baseURL and simple proxy server to transparently inject API key into all requests. This way, if Jupyterlite is part of a multi-user deployment (e.g. university), a single API key could be used for all requests, without users needing to enter own keys. Is there any way to accomplish this currently?

jtpio:
> Is there any way to accomplish this currently?

[...]

PR #179, opened 2025-10-07
Allow specifying an API key for the generic OpenAI compatible provider

jtpio:
Fixes https://github.com/jupyterlite/ai/issues/64

Some providers follow the OpenAI API and would be usable from `jupyterlite-ai` without adding explicit support

- [x] Show a field to provide an optional API key

image

### DIFF
--- a/src/widgets/provider-config-dialog.tsx
+++ b/src/widgets/provider-config-dialog.tsx
@@ -218,34 +218,35 @@ export const ProviderConfigDialog: React.FC<IProviderConfigDialogProps> = ({
             </FormControl>
           )}
 
-          {selectedProvider && selectedProvider?.apiKeyRequirement !== 'none' && (
-            <TextField
-              fullWidth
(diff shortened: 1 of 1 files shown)


src/widgets/provider-config-dialog.tsx
```diff
@@ -219,30 +218,32 @@ export const ProviderConfigDialog: React.FC<IProviderConfigDialogProps> = ({
...
-                      edge="end"
-                    >
-                      {showApiKey ? <VisibilityOff /> : <Visibility />}
-                    </IconButton>
-                  </InputAdornment>
-                )
-              }}
-            />
-          )}
+          <TextField
```
brichet, line 221: Wonder if we should keep `requiredApiKey` and `optionalApiKey` or `undefined`, to avoid confusion with provider that do not use the API key (e.g. `Ollama`, `ChromeAI` or `webLLM`).

PR #195, opened 2025-10-15
Support discovering commands with a query

jtpio:
Implement support for discovering commands using a query argument. This helps reduce the number of commands returned by the tool and the overall token usage, as well as reducing the amount of data sent to the LLM as part of the prompt.

- [x] Support `query`
- [x] Adjust the system prompt accordingly to mention the use of this query argument
- [x] UI tests

image


brichet:
Some thoughts for follow up PRs:
- handle multi words query
- handle ordering, e.g. number of words found or in which property the word(s) has been found (not sure it would be better than letting the LLM find the best though)
- wonder also about an another query argument that would look in the described by property, to specify the commands that can be used or not, depending on the args available (e.g. commands on "notebook" allowing the arg "path"). Again, not sure if the LLM would not handle it better than a filter.

Issue #170, opened 2025-09-29
Cell attachments not sent properly?


jtpio:
## Description

It seems that cell attachments are not correctly sent to the model currently.

## Reproduce

Drag and drop a cell to the chat and ask something about it:

<img width="2250" height="1596" alt="Image" src="https://github.com/user-attachments/assets/84b04db4-d25f-4edd-8598-1273bea44ba9" />

## Expected behavior

When a cell is dragged and dropped into the chat, it should be sent along with the prompt.

## Context

`jupyterlite-ai==0.9.0a2`

nakul-py:
How should i reproduce this bug as i cannot able to drag and drop files or cells into the chat.

jtpio:
Yes for sure.

Locally you can checkout the `next` branch with `git checkout next`.

Then create a new branch as usual, and open a PR against `next` on GitHub.

jtpio:
cc @brichet maybe we could soon merge `next` into `main` and continue working from `main`. Maybe re-enabling the UI tests.

brichet:
> cc [@brichet](https://github.com/brichet) maybe we could soon merge `next` into `main` and continue working from `main`? Maybe re-enabling the UI tests.

Yes, it should be nice to re-enable some tests before. I'll take a look at it as soon as possible.

PR #196, opened 2025-10-15
Fixing clear attachments bug

nakul-py:
Now i think sending attachments works and looks good.

https://github.com/user-attachments/assets/54de64e7-a116-4330-9877-a42de0c606d4


nakul-py:
> As far as I understand, this PR removes the attachment from input sooner than before, and does not wait anymore for the LLM to finalize it answer before clearing. Was that the bug ?

Yes this PR immediate removes the attachment from chat input if sended. This is the bug i have shared in [comment](https://github.com/jupyterlite/ai/issues/170#issuecomment-3405959201)

PR #198, opened 2025-10-15
Better format tool calls that need approval

jtpio:
Improve the formatting of tool calls.

## Single tool call

| Before | After |
|---|---|
| image | image |

---

## Multiple tool calls

| Before | After |
|---|---|
| image | image |

### DIFF
--- a/src/agent.ts
+++ b/src/agent.ts
@@ -719,23 +719,19 @@ export class AgentManager {
   }
 
   /**
-   * Formats tool input for display, handling both objects and pre-stringified JSON.
-   * @param input The tool input to format (object or string)
+   * Formats tool input for display by pretty-printing JSON strings.
+   * @param input The tool input string to format
    * @returns Pretty-printed JSON string
    */
-  private _formatToolInput(input: string | object): string {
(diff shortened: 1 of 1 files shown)


brichet (commented):
Thanks @jtpio, it works well on my side.
I just have a comment, mostly about consistency.

src/agent.ts
```diff
@@ -726,19 +742,13 @@ export class AgentManager {
...
-      parsedToolInput = {};
-    }
 
     this._agentEvent.emit({
       type: 'tool_call_start',
       data: {
         callId: toolCallId,
         toolName,
-        input: parsedToolInput
+        input: this._formatToolInput(toolInput)
```
brichet: Thanks @jtpio.

For the record, it seems that in this case the `toolInput` type is any, so calling the `_formatToolInput` could raise an error.
But we can probably try like this and see if it fails sometime.

jtpio: There are a couple of leftover `any` in the code base from the early iteration, but we should indeed get rid of them.

Will take a look to clean that up.

PR #200, opened 2025-10-15
Move the secrets manager toggle below the providers

jtpio:
Move the secret manager below the list of providers, because it happens to be quite easy to misclick it currently, which results in clearing API keys currently stored in the settings and having to reconfigure them.

image


brichet (approved):
Thanks @jtpio, looks good.

I'll check the behavior when clicking, I though the keys wasn't lost, but copied from the settings to secrets manager and vice versa.

jtpio:
> I'll check the behavior when clicking, I though the keys wasn't lost, but copied from the settings to secrets manager and vice versa.

It could be because of the placeholder:

https://github.com/jupyterlite/ai/blob/2be02bc2e6562fdd837424ed442158d75401d0df/src/tokens.ts#L229

Issue #199, opened 2025-10-15
Expose the prompt used for inline completion


jtpio:
### Problem

Currently, we expose the system prompt used by the agent in one of the tabs of the AI Settings panel. However, the one used for the inline completion is not exposed.

<img width="3014" height="2498" alt="Image" src="https://github.com/user-attachments/assets/fc28d4cb-65eb-448e-a735-26d1d4446a69" />

### Proposed Solution

Expose both prompts so the end users could play around with them and also tweak them:

https://github.com/jupyterlite/ai/blob/f3b1c6a6a2693e792307280b28bd312528b166da/src/completion/completion-provider.ts#L36-L47

### Additional context

`jupyterlite-ai==0.9.0a3`

PR #204, opened 2025-10-17
Debounce settings updates when editing the system prompt

jtpio:
Previously editing the prompt was triggering requests to update the settings on each keystroke.

Now:

https://github.com/user-attachments/assets/db9e3ff5-c3af-4668-997e-b3cdc83b497e


brichet:
Thanks @jtpio, it looks great.

We'll need to use the same mechanism for the completion prompt, after #199.
We can probably use the same debouncer with both values, since `model.updateConfig()` checks if the value has changed before updating.

brichet (approved):
Thanks @jtpio, LGTM

PR #209, opened 2025-10-20
Making Ai Agent able to Reading content for attached cells

### LINKED ISSUE #170: Cell attachments not sent properly?

nakul-py:
Now our Ai Agent can easily able to read content of any notebook cells attached in chat. 

### Before

In this screen recording we can see as i share 3rd cell but we get colors list of 1st cell.

https://github.com/user-attachments/assets/44e0f898-ccc1-42e5-a3f6-b5c2035e2d21

### After

In this we can see i am sharing 3rd cell and get color list of same 3rd cell.

https://github.com/user-attachments/assets/4240f764-2fcc-42be-aca6-3219b1d8635d

- Closes #170

CC @jtpio and @brichet

### DIFF
--- a/src/chat-model.ts
+++ b/src/chat-model.ts
@@ -626,6 +626,11 @@ ${toolsList}
         return null;
       }
 
+      const kernelLang =
+        model.content?.metadata?.language_info?.name ||
+        model.content?.metadata?.kernelspec?.language ||
+        'text';
+
       const selectedCells = attachment.cells
         .map(cellInfo => {
           const cell = model.content.cells.find(
@@ -637,7 +642,9 @@ ${toolsList}
 
           const code = cell.source || '';
           const cellType = cell.cell_type;
-          return `**Cell [${cellInfo.id}] (${cellType}):**\n\`\`\`${cellType === 'code' ? 'python' : ''}\n${code}\n\`\`\``;
+          const lang = cellType === 'code' ? kernelLang : cellType;
+
+          return `**Cell [${cellInfo.id}] (${cellType}):**\n\`\`\`${lang}\n${code}\n\`\`\``;
         })
         .filter(Boolean)
         .join('\n\n');


brichet:
@nakul-py thanks for working on this.

It does not work on my side, there is no cell contents attached when I check the request, only `-- Attached Files ---**Notebook: test.ipynb**`.
In the example that you shared, the LLM uses tools, so we can't make sure that the cell content is attached.

brichet:
> it works on my side expect request info.

Indeed, Sorry about that, I probably made a mistake with my environment. It works now after rebuilding.

src/chat-model.ts
```diff
@@ -598,6 +606,52 @@ ${toolsList}
...
+          const cell = model.content.cells.find(
+            (c: any) => c.id === cellInfo.id
+          );
+          if (!cell) {
+            return null;
+          }
+
+          const code = cell.source || '';
+          const cellType = cell.cell_type;
+          return `**Cell [${cellInfo.id}] (${cellType}):**\n\`\`\`${cellType === 'code' ? 'python' : ''}\n${code}\n\`\`\``;
```
brichet: We should not assume that the content is Python, it can be any other language.
We could either get the information from the Notebook metadata, or let the LLM find out.

brichet (commented):
Thanks @nakul-py, it works well.

src/chat-model.ts
nakul-py: Yes You're right! I've implemented this in my upcoming commit. :)

brichet (approved):
Thanks @nakul-py, looks nice !

src/chat-model.ts
```diff
@@ -598,6 +606,59 @@ ${toolsList}
...
+      const selectedCells = attachment.cells
+        .map(cellInfo => {
+          const cell = model.content.cells.find(
+            (c: any) => c.id === cellInfo.id
+          );
+          if (!cell) {
+            return null;
+          }
+
+          const code = cell.source || '';
```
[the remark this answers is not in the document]
jtpio: Opened https://github.com/jupyterlite/ai/issues/215

PR #213, opened 2025-10-23
Split and unified cell diffs

jtpio:
Various improvements to cell diffs:

- [x] Update to the new release `jupyterlab-diff`: https://github.com/jupyter-ai-contrib/jupyterlab-diff/releases/tag/v0.6.0
- [x] Add new options to configure cell diffs (split or unified)

https://github.com/user-attachments/assets/1215ec7c-d016-42f3-8da2-2f4ae53a1224

### DIFF
--- a/src/diff-manager.ts
+++ b/src/diff-manager.ts
@@ -1,6 +1,6 @@
 import { CommandRegistry } from '@lumino/commands';
-import { IDiffManager } from './tokens';
 import { AISettingsModel } from './models/settings-model';
+import { IDiffManager, IShowCellDiffParams } from './tokens';
 
 /**
  * Command IDs for unified cell diffs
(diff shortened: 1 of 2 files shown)


brichet (approved):
Thanks @jtpio, this looks great.

My only concern was about the overlapping of the cell toolbar when using the unified diff, cross reference to https://github.com/jupyter-ai-contrib/jupyterlab-diff/issues/22

jtpio:
> My only concern was about the overlapping of the cell toolbar when using the unified diff, cross reference to [jupyter-ai-contrib/jupyterlab-diff#22](https://github.com/jupyter-ai-contrib/jupyterlab-diff/issues/22)

Yeah not sure yet how to best handle that. In the worst case users can still use the split (side-by-side) view if they prefer, or just "Accept all" / "Reject all".

PR #220, opened 2025-11-04
Fix the type of response for the generic provider

jtpio:
Restore part of https://github.com/jupyterlite/ai/pull/171, which seems to have been lost during https://github.com/jupyterlite/ai/pull/191


jtpio:
> I wonder if we could/should use the `openai-compatible` provider instead of `openai` directly for the generic provider. https://ai-sdk.dev/providers/openai-compatible-providers

Probably it would make sense yes, if it uses the `/chat/completions` endpoint by default (or if it's possible to use it like with the current use of `openai()`)

PR #221, opened 2025-11-06
Use OpenAI compatible provider

brichet:
This PR make use of `openai-compatible` mrovider from `ai-sdk` for the generic provider, instead `openai` provider.

It also add tests using the generic provider.

Related to comments in #220

> I wonder if we could/should use the `openai-compatible` provider instead of `openai` directly for the generic provider.
https://ai-sdk.dev/providers/openai-compatible-providers

> Probably it would make sense yes, if it uses the `/chat/completions` endpoint by default (or if it's possible to use it like with the current use of `openai()`)

### DIFF
--- a/ui-tests/tests/code-completion.spec.ts
+++ b/ui-tests/tests/code-completion.spec.ts
@@ -34,8 +34,6 @@ TEST_PROVIDERS.forEach(({ name, settings }) =>
       page.on('request', data => {
         if (data.method() === 'POST') {
           const url = new URL(data.url());
-          console.log('hostname', url.hostname);
-          console.log('pathname', url.pathname);
           if (
             ['127.0.0.1', 'localhost'].includes(url.hostname) &&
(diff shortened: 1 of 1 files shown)


ui-tests/tests/code-completion.spec.ts
```diff
@@ -4,70 +4,78 @@
...
-  // Retry by typing more content if ghost text doesn't appear
-  const ghostText = cell!.locator('.jp-GhostText');
-  const maxRetries = 3;
-  let retries = 0;
-  let ghostTextVisible = false;
+      page.on('request', data => {
+        if (data.method() === 'POST') {
+          const url = new URL(data.url());
+          console.log('hostname', url.hostname);
+          console.log('pathname', url.pathname);
```
jtpio: Looks like these logs could be removed?

jtpio (approved):
Looks good, thanks!

Just left a minor comment.

PR #226, opened 2025-11-12
Adding Cell Outputs as a part of attachments

### LINKED ISSUE #215: Include the cell outputs as part of attachments
Not necessarily in this PR, but as a follow-up it could be interesting to also include the cell outputs, so users could ask the agent to fix errors.

_Originally posted by @jtpio in https://github.com/jupyterlite/ai/pull/209#discussion_r2460280053_

nakul-py:
## Description
Including the cell outputs as part of attachments so users can ask the agent to fix errors and update code now.

### Screen Recording

https://github.com/user-attachments/assets/02680952-053e-4a7c-bcb9-1273fed32867

----
- Closes #215 

cc @jtpio @brichet

### DIFF
--- a/src/chat-model.ts
+++ b/src/chat-model.ts
@@ -727,7 +727,12 @@ ${toolsList}
                   if (!data) {
                     return '';
                   }
-                  return extractDisplay(data);
+                  try {
+                    return extractDisplay(data);
+                  } catch (e) {
+                    console.error('Cannot extract cell output', e);
+                    return '';
+                  }
                 }
                 return '';
               })


src/chat-model.ts
```diff
@@ -646,7 +646,45 @@ ${toolsList}
...
+                  traceback: any[];
+                  data: any;
+                }) => {
+                  if (output.output_type === 'stream') {
+                    return output.text;
+                  } else if (output.output_type === 'error') {
+                    return output.traceback.join('\n');
+                  } else if (
+                    output.output_type === 'execute_result' ||
+                    output.output_type === 'display_data'
```
brichet: I wonder if there are use cases for it. Testing it with `ipywidgets`, the output is truncated very early:
For an `Image` widget for example, the output is `Image(value=b'\xff\xd8\xff\xe1\x15\xe6Exif\x00\x00MM\x00*\x00\x00\x00\x08\x00\x10\x01\x00\x00\x03\x00\x00\x00\…`

src/chat-model.ts
```diff
@@ -646,7 +646,45 @@ ${toolsList}
...
-          return `**Cell [${cellInfo.id}] (${cellType}):**\n\`\`\`${lang}\n${code}\n\`\`\``;
+          let outputs = '';
+          if (cellType === 'code' && Array.isArray(cell.outputs)) {
+            outputs = cell.outputs
+              .map(
+                (output: {
+                  output_type: string;
+                  text: any;
+                  traceback: any[];
+                  data: any;
```
brichet: Is there a way to get real types ? There are a lot of `any` type here, the cell itself is `any`.

nakul-py:
> Probably not for this PR, but I also wonder if we can (and should) get the model from the widget, instead of using the `documentManager`
> 
> https://github.com/jupyterlite/ai/blob/06a3df8e4d2c304fbf31448489b09e90bc781239/src/chat-model.ts#L624
> 
> .
> If the document is not saved, the cell sent to the LLM can be different from the one in the widget (as well as the whole Notebook, BTW), which can be very confusing.

Yes its a good idea and future proof also.:)

src/chat-model.ts
```diff
@@ -646,11 +648,65 @@ ${toolsList}
...
+          return (
+            `**Cell [${cellInfo.id}] (${cellType}):**\n` +
+            `\`\`\`${lang}\n${code}\n\`\`\`` +
+            (outputs ? `\n**Outputs:**\n\`\`\`text\n${outputs}\n\`\`\`` : '')
+          );
         })
         .filter(Boolean)
         .join('\n\n');
 
+      console.log('Selected notebook cells content:', selectedCells);
```
brichet: We should remove this log.

nakul-py: Ohh! i forgot to remove this debugger log.

src/chat-model.ts
```diff
@@ -646,11 +648,65 @@ ${toolsList}
...
+                    data['application/vnd.jupyter.widget-view+json'];
+                  if (
+                    widgetData &&
+                    typeof widgetData === 'object' &&
+                    'model_id' in widgetData
+                  ) {
+                    const modelId = (widgetData as { model_id?: string })
+                      .model_id;
+                    return `Widget: ${modelId ?? 'unknown model'}`;
+                  }
```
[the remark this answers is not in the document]
nakul-py: @brichet and @jtpio I think we can make a array for all usecases like 
```ts
 const DISPLAY_PRIORITY = [
            'application/vnd.jupyter.widget-view+json',
            'application/javascript',
            'text/html',
            'image/svg+xml',
            'image/png',
            'image/jpeg',
            'text/markdown',
            'text/latex',
            'text/plain'
          ];
```

and extract the one display that is needed.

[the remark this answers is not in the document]
nakul-py: Here's how it works now (also updated PR's description video)

https://github.com/user-attachments/assets/32a34b51-b3d4-4a9c-be51-2db38340cf2f

src/chat-model.ts
```diff
@@ -646,7 +648,102 @@ ${toolsList}
...
+                  return `${err.ename}: ${err.evalue}\n${(err.traceback || []).join('\n')}`;
+                } else if (
+                  output.output_type === 'execute_result' ||
+                  output.output_type === 'display_data'
+                ) {
+                  const data = (output as nbformat.IDisplayData).data;
+                  if (!data) {
+                    return '';
+                  }
+                  return extractDisplay(data);
```
```suggestion
                  try {
                    return extractDisplay(data);
                  } catch (e) {
                    console.error('Cannot extract cell output', e);
                    return '';
                  }
```

brichet (commented):
Thanks @nakul-py, I have one last suggestion to avoid error, and I think we should be able to merge it after.

src/chat-model.ts
nakul-py: Done. Thanks @brichet 🧡

Issue #182, opened 2025-10-09
Drop the Ollama built-in provider?


jtpio:
### Problem

Currently `jupyterlite-ai` includes explicit and built-in support for Ollama.

However Ollama is compatible with the OpenAI API: https://ollama.com/blog/openai-compatibility

So maybe it could make sense to drop support for Ollama as built-in-provider, and instead document how to use the generic OpenAI-compatible provider?

This would also mean dropping the dependency on the third-party provider: https://github.com/jupyterlite/ai/blob/f9186f0a1aacba2195be4b5e1aeb3d15054201a4/package.json#L86

### Proposed Solution

<img width="1788" height="1150" alt="Image" src="https://github.com/user-attachments/assets/92f4291a-0378-4819-a1d7-66d745adfb7c" />

### Additional context

The main downside would be to manually provide the base URL for Ollama when using the generic provider.

Unless the UI could suggest some common URLs, for example Ollama, Openrouter and LiteLLM proxy.

brichet:
There is now a test that use the generic provider with Ollama, so yes we can probably remove it now.

I opened #230 to add some URL suggestions in the provider info.

PR #230, opened 2025-11-12
Add optional URL suggestions to provider info

brichet:
Adds optional URL suggestions to the provider info, and display it as an autocomplete dropdown list when setting the provider.

Screenshot from 2025-11-12 20-51-01

Related to #182

### DIFF
--- a/src/providers/built-in-providers.ts
+++ b/src/providers/built-in-providers.ts
@@ -212,7 +212,7 @@ export const ollamaProvider: IProviderInfo = {
   apiKeyRequirement: 'none',
   defaultModels: [],
   supportsBaseURL: true,
-  URLS: [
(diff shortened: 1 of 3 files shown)


src/providers/built-in-providers.ts
```diff
@@ -212,7 +212,12 @@ export const ollamaProvider: IProviderInfo = {
   apiKeyRequirement: 'none',
   defaultModels: [],
   supportsBaseURL: true,
-  supportsHeaders: true,
+  URLS: [
```
jtpio: Wondering if this should be named `baseUrls` to be consistent with the user facing string?

brichet: Right, I updated it.

PR #233, opened 2025-11-15
Use the generic provider for Ollama

jtpio:
Fixes https://github.com/jupyterlite/ai/issues/182


README.md
```diff
@@ -52,6 +52,18 @@ The process is different for each provider, so you may refer to their documentat
 
 ![screenshot showing the dialog to add a new provider](https://github.com/user-attachments/assets/823c71c6-5807-44c8-80b6-2e59379a65d5)
 
+### Using a generic OpenAI-compatible provider
```
brichet: I wonder if the sections `Using Ollama` and `Using LiteLLM Proxy` should be under this one. Currently they are at the same level.

PR #237, opened 2025-11-24
Getting model from widget first than fallback to documentManager

### LINKED ISSUE #234: Get the model from the widget, instead of  the `documentManager`
### Problem
If the document is not saved, the cell sent to the LLM can be different from the one in the widget (as well as the whole Notebook, BTW), which can be very confusing.

### Proposed Solution

Get the model from the widget, instead of using the `documentManager`
### Additional context

Raised in https://github.com/jupyterlite/ai/pull/226#issuecomment-3522755759

nakul-py:
First try to getting model from `widget` than fallback to `documentManager` for both files and notebook cells.

https://github.com/user-attachments/assets/f089aaf4-5457-499b-ada4-c65b99ea48d0

https://github.com/user-attachments/assets/8ffd1aa7-c656-46de-ac78-048e9119a185

- Closes #234 

cc @jtpio @brichet

### DIFF
--- a/src/chat-model.ts
+++ b/src/chat-model.ts
@@ -633,10 +633,10 @@ ${toolsList}
       const widget = this.input.documentManager?.findWidget(
         attachment.value
       ) as IDocumentWidget<Notebook, INotebookModel> | undefined;
-      let cellData: nbformat.ICell[] | null = null;
+      let cellData: nbformat.ICell[];
       let kernelLang = 'text';
 
-      const ymodel = widget?.context?.model?.sharedModel as YNotebook;
+      const ymodel = widget?.context.model.sharedModel as YNotebook;
 
       if (ymodel) {
         const nb = ymodel.toJSON();
@@ -644,8 +644,8 @@ ${toolsList}
         cellData = nb.cells;
 
         const lang =
-          nb.metadata?.language_info?.name ||
-          nb.metadata?.kernelspec?.language ||
+          nb.metadata.language_info?.name ||
+          nb.metadata.kernelspec?.language ||
           'text';
 
         kernelLang = String(lang);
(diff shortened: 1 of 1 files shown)


brichet:
Thanks @nakul-py for working in this.

After a quick test, it works well on my side with cell but with a notebook I have something like 
  ```
  **File: blabla.ipynb**```json[object Object]```
  ```


Can you use the types of objects instead of `any` ? That would probably help preventing unexpected errors, and also simplify review.

src/chat-model.ts
```diff
@@ -623,23 +623,39 @@ ${toolsList}
...
 
     try {
-      const model = await this.input.documentManager?.services.contents.get(
-        attachment.value
-      );
-      if (!model || model.type !== 'notebook') {
-        return null;
-      }
+      // Try reading from live notebook if open
+      const widget = this.input.documentManager?.findWidget(attachment.value);
```
brichet: As far as I know, the widget can only be a `IDocumentWidget<Notebook, INotebookModel> | undefined`. 
Casting it would avoid typing the following `sharedModel` and `cells` as `any`.
Did you see use case where the widget is of another type ?

src/chat-model.ts
```diff
@@ -623,23 +629,45 @@ ${toolsList}
...
+          model.content?.metadata?.kernelspec?.language ||
+          'text';
+      }
 
       const selectedCells = attachment.cells
         .map(cellInfo => {
-          const cell = model.content.cells.find(
-            (c: any) => c.id === cellInfo.id
-          );
+          const cell = cellData!.find(c => c.id === cellInfo.id);
```
```suggestion
          const cell = cellData.find(c => c.id === cellInfo.id);
```

*Related to above comments*

src/chat-model.ts
```diff
@@ -777,36 +806,63 @@ ${toolsList}
...
+        const ymodel = model.sharedModel as YNotebook;
+
+        if (typeof ymodel.getSource === 'function') {
+          const source = ymodel.getSource();
+          return typeof source === 'string'
+            ? source
+            : JSON.stringify(source, null, 2);
+        }
+
+        if (typeof ymodel.toJSON === 'function') {
```
brichet: Can we reach this comparison ?
The `YNotebook` always has a `getSource()` function defined, AFAIK https://github.com/jupyter-server/jupyter_ydoc/blob/fdbf658e072467f49048ddef513a5e569207f25b/javascript/src/ynotebook.ts#L408

brichet (commented):
Thanks @nakul-py  for updating the PR :+1: 

I have some nitpick comments

nakul-py:
@brichet Now everything looks fine. 👍

PR #244, opened 2025-12-11
Add favicon to the docs

jtpio:
**Before**

image

**After**

image

### DIFF
--- a/myst.yml
+++ b/myst.yml
@@ -34,5 +34,5 @@ site:
   options:
     logo: style/icons/jupyternaut-lite.svg
     logo_text: JupyterLite AI
-    favicon: style/icons/jupyternaut-lite.svg
+    favicon: docs/favicon.ico
     style: docs/custom.css


myst.yml
```diff
@@ -32,4 +32,7 @@ project:
 site:
   template: book-theme
   options:
+    logo: style/icons/jupyternaut-lite.svg
     logo_text: JupyterLite AI
+    favicon: docs/favicon.ico
```
brichet: Can't we use the same file for both favicon and logo, as you suggest @jtpio ?

It seems to work well locally.

jtpio: Yes let's do that.

jtpio: Hmm it seems the favicon is not appearing anymore on the preview:

image

jtpio: Looks like it's still `favicon.ico` being requested:

image

If it's too complicated to configure with myst, let's just revert to what we had before with the extra `.ico` file?

PR #247, opened 2026-01-13
Align deps on stable releases

jtpio:
Mechanical bump to align on stable releases.

### DIFF
--- a/demo/uv.lock
+++ b/demo/uv.lock
@@ -838,7 +838,7 @@ wheels = [
 
 [[package]]
 name = "jupyterlab"
-version = "4.5.0b1"
+version = "4.5.2"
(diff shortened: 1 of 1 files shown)


brichet:
I wonder if it worth deploying the demo in the doc, or increasing the built artifact https://github.com/actions/upload-pages-artifact?tab=readme-ov-file#inputs- (default to 1 day).

jtpio:
> I wonder if it worth deploying the demo in the doc, or increasing the built artifact https://github.com/actions/upload-pages-artifact?tab=readme-ov-file#inputs- (default to 1 day).

The demo in docs would indeed be useful to more easily try PRs. Maybe we can do both.

PR #250, opened 2026-01-19
Use the functiongemma model for tool call tests

brichet:
Attempt to have more reliable tests on tool call by using the functiongemma model.
This PR also replace the use of [pydantic/ollama-action](https://github.com/pydantic/ollama-action) with [ai-action/ollama-action](https://github.com/ai-action/ollama-action/tree/master), which seems more active.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -37,7 +37,7 @@ jobs:
       - run: ollama --version
         shell: bash
 
-      - name: Cache model
+      - name: Cache Ollama models
         id: cache-ollama-models
         uses: actions/cache@v5
         with:
@@ -45,7 +45,7 @@ jobs:
           key: ollama-cache-2026-01-20
 
       - name: Install qwen2.5:0.5b and functiongemma model
-        if: steps.cache-ollama-model.outputs.cache-hit != 'true'
+        if: ${{ steps.cache-ollama-models.outputs.cache-hit != 'true' }}
         run: |
           ollama pull qwen2.5:0.5b
           ollama pull functiongemma


.github/workflows/tests.yml
```diff
@@ -31,11 +31,16 @@ jobs:
           pip install "jupyterlab>=4.5.0"
           pip install -e ".[test]"
 
-      - name: Install Ollama with a small model
-        uses: pydantic/ollama-action@v3
+      - name: Install Ollama with qwen2.5:0.5b model
+        uses: ai-action/ollama-action@v2
```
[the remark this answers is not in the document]
nakul-py: > Another option could be to pin on a specific hash.

I am agree with this.

.github/workflows/tests.yml
```diff
@@ -31,11 +31,16 @@ jobs:
           pip install "jupyterlab>=4.5.0"
           pip install -e ".[test]"
 
-      - name: Install Ollama with a small model
-        uses: pydantic/ollama-action@v3
+      - name: Install Ollama with qwen2.5:0.5b model
+        uses: ai-action/ollama-action@be18a6c5c9977cc0510d1d705edd02d574269c93
```
jtpio: Actually this action seems to depend on another `ai-action/ollama-action@v2` here: https://github.com/ai-action/ollama-action/blob/be18a6c5c9977cc0510d1d705edd02d574269c93/action.yml#L30

So not sure pinning helps if this inner action gets updated?

nakul-py: Nice catch @jtpio if the action itself still pulls other actions by tag, then pinning here doesn’t fully address the supply-chain concern.

I’m fine reverting to `pydantic/ollama-action` and only adjusting the model if that’s the main source of flakiness.

Issue #253, opened 2026-01-27
Support skills


jtpio:
Skills are relatively new but already supported by many CLI tools and IDEs: https://agentskills.io/home

Would be great to support that in jupyterlite-ai too.

So end users could at least augment their system prompts and general instructions with markdown files (to start with), that would then be discovered by the model.

PR #259, opened 2026-02-02
Support skills

jtpio:
Fixes https://github.com/jupyterlite/ai/issues/253

Aligning with https://agentskills.io.

This page includes a table comparing skills to MCP: https://claude.com/blog/extending-claude-capabilities-with-skills-mcp-servers

- [x] Support collecting skills from a `.agents/skills`, since this convention seems to be adopted by other tools already like Codex or opencode 
- [x] Add `ISkillRegistry`
- [x] List available skills with `/skills`
- [x] Respect progressive disclosure
- [x] Documentation
- [x] UI test

**Basic test skill**

https://github.com/user-attachments/assets/e04022da-cda6-4151-abef-f73ab4e07aaf

**Bootstrap workflow**

https://github.com/user-attachments/assets/7c058a85-c625-493f-a5ad-5329830c43d8

### DIFF
--- a/src/skills/skill-loader.ts
+++ b/src/skills/skill-loader.ts
@@ -4,6 +4,7 @@
  */
 
 import { Contents } from '@jupyterlab/services';
+import { PathExt } from '@jupyterlab/coreutils';
 
 import { parseSkillMd, IParsedSkill } from './parse-skill';
 
@@ -86,7 +87,7 @@ async function loadSkills(
     if (!children.some(c => c.type === 'directory' && c.name === segment)) {
       return skills;
     }
-    currentPath = currentPath ? `${currentPath}/${segment}` : segment;
+    currentPath = PathExt.join(currentPath, segment);
   }
 
   const dirModel = await contentsManager.get(skillsPath, { content: true });
@@ -163,7 +164,7 @@ async function collectResourcePaths(
         await walk(item.path);
       } else if (item.type === 'file' && item.name !== 'SKILL.md') {
         // Store path relative to the skill directory
-        const relativePath = item.path.slice(basePath.length + 1);
+        const relativePath = PathExt.relative(basePath, item.path);
         resourcePaths.push(relativePath);
       }
     }


jtpio:
>  Commands prefixed with skills: are treated as skills

Thinking more about this, maybe we should actually have a proper skill registry to make it easier for extensions to register skills, instead of relying on naming convention with the command system.

jtpio:
@brichet if you would like to have another look, the first two points should be addressed now. We can look into the skill bootstrap command separately maybe.

src/skills/skill-loader.ts
```diff
@@ -0,0 +1,174 @@
...
+ * @param skillsPath - Path to the skills directory (e.g. ".agents/skills")
+ * @returns Array of loaded skill definitions
+ */
+async function loadSkills(
+  contentsManager: Contents.IManager,
+  skillsPath: string
+): Promise<ISkillFileDefinition[]> {
+  const skills: ISkillFileDefinition[] = [];
+
+  // Walk each path segment from root to verify the directory exists before fetching it.
```
brichet: Is this to avoid server error if path doesn't exist ?

src/skills/skill-loader.ts
```diff
@@ -0,0 +1,174 @@
...
+    for (const item of dirModel.content as Contents.IModel[]) {
+      // Skip checkpoint directories
+      if (item.name === '.ipynb_checkpoints') {
+        continue;
+      }
+      if (item.type === 'directory') {
+        await walk(item.path);
+      } else if (item.type === 'file' && item.name !== 'SKILL.md') {
+        // Store path relative to the skill directory
+        const relativePath = item.path.slice(basePath.length + 1);
```
brichet: Should we use `PathExt` here too.

src/agent.ts
```diff
@@ -597,59 +660,108 @@ export class AgentManager {
...
-        temperature,
-        maxOutputTokens: maxTokens,
-        stopWhen: stepCountIs(maxTurns)
-      });
-    } catch (error) {
-      console.warn('Failed to initialize agent:', error);
+  /**
+   * Rebuild the agent using cached resources and the current skills snapshot.
+   */
+  private _rebuildAgent(): void {
```
brichet: Wonder if we should open an issue on ai-sdk to allow updating the `instructions`, that would simplify a lot the agent manager and factory implementation.

brichet (approved):
Thanks @jtpio, it's really nice.

I just have a few comments but nothing required.

src/skills/skill-loader.ts
jtpio: Yes to avoid potential 404 in the dev tools console mostly.

Otherwise we could have simplified to mostly one request.

src/agent.ts
jtpio: Another approach would have been to treat skills (and potentially other dynamic information) as a separate system message, that could be added to the chat history after the system prompt. Although it would use a bit more tokens and still require keeping track of skills updates.

PR #262, opened 2026-02-06
Show context window usage

jtpio:
Display the context window usage to have a better idea of how much space there is and give a better hint of the current history size.

- [x] Get context window information from https://models.dev
- [x] Run the script locally to update the information about context windows
- [x] New setting to show / hide the context window usage 

image

image


jtpio:
Maybe we can get some information from https://github.com/anomalyco/models.dev via a custom script to be triggered manually when updating the list of models.

jtpio:
Maybe it's good enough and will already help users have a better visibility on how much of the context window is being used.

As a follow-up, maybe we could also look into getting information from https://models.dev to manage the list of default built-in model ids (still as an offline manual sync step, not at runtime).

src/components/usage-display.tsx
brichet, line 1: For the sake of consistency, we could probably rename this file and this class since it now displays token usage and context usage.

src/components/token-usage-display.tsx
```diff
@@ -56,51 +75,95 @@ export const TokenUsageDisplay: React.FC<ITokenUsageDisplayProps> = ({
               }
 
               const total = tokenUsage.inputTokens + tokenUsage.outputTokens;
-              if (total === 0) {
+              const hasTokenUsage = showTokenUsage && total > 0;
```
brichet: We should show token usage even if it 0. That can be confusing to expect it but having it empty.

src/components/token-usage-display.tsx
```diff
@@ -56,51 +75,95 @@ export const TokenUsageDisplay: React.FC<ITokenUsageDisplayProps> = ({
...
+              const hasKnownContextWindow =
+                showContextUsage && tokenUsage.contextWindow !== undefined;
+              const hasContextEstimate =
+                hasKnownContextWindow &&
+                tokenUsage.contextUsagePercent !== undefined &&
+                tokenUsage.lastRequestInputTokens !== undefined;
+              const shouldShowContextBadge = showContextUsage;
```
brichet: Is it necessary to have a new constant ?

src/widgets/provider-config-dialog.tsx
```diff
@@ -676,6 +681,41 @@ export const ProviderConfigDialog: React.FC<IProviderConfigDialogProps> = ({
                   inputProps={{ min: 1, max: 100 }}
                 />
 
+                <TextField
+                  fullWidth
+                  label={trans.__('Context Window (Optional)')}
+                  type="number"
+                  value={parameters.contextWindow ?? ''}
```
brichet, line 688: Should we fill it with the default value, since value is used to compute the percentage ?

src/widgets/provider-config-dialog.tsx
```diff
@@ -676,6 +681,41 @@ export const ProviderConfigDialog: React.FC<IProviderConfigDialogProps> = ({
...
+                    parameters.contextWindow === undefined
+                      ? trans.__(
+                          'Using provider metadata default of %1 tokens for this model unless you override it here.',
+                          selectedModelInfo.contextWindow.toLocaleString()
+                        )
+                      : trans.__(
+                          'Model context window size in tokens (used for context usage estimation)'
+                        )
+                  }
+                  inputProps={{ min: 1 }}
```
brichet: The `inputProps` property seems to be deprecated.

src/tokens.ts
```diff
@@ -776,6 +791,22 @@ export interface ITokenUsage {
...
+  /**
+   * Configured context window size for the active provider/model.
+   */
+  contextWindow?: number;
+
+  /**
+   * Estimated context usage percentage for the most recent request.
+   */
+  contextUsagePercent?: number;
 }
```
brichet, line 805: I wonder if we need to store all these values. Only the `lastRequestInputTokens` seems to be relevant, the `contextWindow` should be the same as the one from model info, and the `contextUsagePercent` is a simple computation between the 2 others (if I'm not mistaken).

src/tokens.ts
```diff
@@ -204,6 +204,13 @@ export interface IProviderToolCapabilities {
 /**
  * Provider information
  */
+export interface IProviderModelInfo {
```
brichet, line 207: For the record, I wonder if we should include the `webSearch` and `webFetch` capabilities. I think it is also per model capabilities, but i don't know if the list exist somewhere.

brichet (commented):
Thanks @jtpio for working on this.

package.json
```diff
@@ -53,7 +53,8 @@
         "watch:src": "tsc -w --sourceMap",
         "watch:labextension": "jupyter labextension watch .",
         "docs": "jupyter book start",
-        "docs:build": "sed -e 's/\\[@/[/g' -e 's/@/\\&#64;/g' CHANGELOG.md > docs/_changelog_content.md && jupyter book build --html"
+        "docs:build": "sed -e 's/\\[@/[/g' -e 's/@/\\&#64;/g' CHANGELOG.md > docs/_changelog_content.md && jupyter book build --html",
+        "sync:model-context-windows": "node scripts/sync-model-context-windows.mjs && prettier --write src/providers/generated-context-windows.ts && eslint --fix src/providers/generated-context-windows.ts"
```
brichet, line 57: In a follow up PR we could add a workflow to update it every X days and open a PR.

src/providers/generated-context-windows.ts
```diff
@@ -0,0 +1,102 @@
+/**
+ * This file is generated by `jlpm sync:model-context-windows`.
+ * Source: https://models.dev/api.json
+ * Backed by: https://github.com/anomalyco/models.dev
+ * Generated: 2026-04-08T16:23:34.080Z
+ */
+
+import type { IProviderModelInfo } from '../tokens';
+
+export const BUILT_IN_PROVIDER_MODEL_INFO: Record<
```
brichet, line 10: For future PR too, we may want to expose it via a token, to allow extensions adding providers to add context window information.

jtpio:
Thanks @brichet for the review.

Just pushed a commit to simplify things a bit.

src/components/usage-display.tsx
```diff
@@ -75,29 +75,42 @@ export const TokenUsageDisplay: React.FC<ITokenUsageDisplayProps> = ({
               }
 
               const total = tokenUsage.inputTokens + tokenUsage.outputTokens;
-              const hasTokenUsage = showTokenUsage && total > 0;
+              const hasTokenUsage = showTokenUsage;
```
brichet: This constant may not be necessary.

brichet (approved):
Thanks @jtpio.

Previous comments can be addressed in follow up PR(s) if they are relevant.

brichet:
Thanks

PR #268, opened 2026-02-13
Bump @jupyter/chat to 0.20.0-alpha.1

brichet:
Bump `@jupyter/chat` to 0.20.0-alpha.0.

The mime bundle renderer can be tested with the following diff

[diff-mime-bundle.txt](https://github.com/user-attachments/files/25299660/diff-mime-bundle.txt)

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
         "@ai-sdk/mistral": "^3.0.19",
         "@ai-sdk/openai": "^3.0.26",
         "@ai-sdk/openai-compatible": "^2.0.28",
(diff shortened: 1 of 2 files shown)


brichet:
This is probably because of the changes in https://github.com/jupyterlab/jupyter-chat/pull/349/changes#diff-57bf1d7c35b2582f383c7acd1ec828d59794161527370c63feff9ba4560ab128, instead of replacing the content as before, we now remove and add the new content in 2 times.
We need to improve it.

PR #270, opened 2026-02-18
Include lite deployment in doc

brichet:
This PR includes a lite deployment in the documentation, that would help testing changes in a PR.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -13,12 +13,16 @@ build:
 
     # Build JupyterLite assets
     pre_build:
-      - pip install uv
-      - uv python install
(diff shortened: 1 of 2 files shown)


brichet:
> > I wonder if we should keep the `github_pages` deployment.
> 
> The URL may be nicer: https://jupyterlite.github.io/ai/

Yes, the problem is that you can't have several build, so each PR would update the unique deployment.
Wonder if we can use the RTD built one only on PR, and use the github_pages on main...

jtpio:
Ah yes, I meant we can keep the GitHub Pages deployment as the user facing deployment (GitHub URL link and in the README), and the RTD deployment mostly for testing PR changes?

PR #276, opened 2026-02-20
Use jupyter-chat-components for tool call components

brichet:
This PR should not change the experience, but it uses an external mime renderer extension to display tool call components.


jtpio (approved):
Thanks!

jtpio (approved):
Thanks!

Issue #271, opened 2026-02-18
LLM stuck because of message history


brichet:
It seems the LLM can be stuck because of a wrong history.
This has been observed with Mistral model (ministral-8b-3)

## Description

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/965bb6cd-2417-4efe-8744-940b3917b352" />

It seems that it occurs if a user message is inserted before a tool call. The request response displays:
```json
{
  "object":"error",
  "message":"Unexpected role 'tool' after role 'user'",
  "type":"invalid_request_message_order",
  "param":null,
  "code":"3230"
}
```

## Reproduce

Maybe sending a message while the LLM is using tools ?

## Expected behavior

The LLM to be available, even if an error occurred.

## Context

Happened with Mistral AI model. Not sure if it occurs with other LLMs. 

cc. @trungleduc

Yahiewi:
Hey @brichet, I've started to look into this issue and here's how I've been able to reliably reproduce it:
-Create a notebook with `time.sleep(10)`.
-Ask the agent to run it.
-While the cell is being run and the agent is waiting for the result of the `run-cell` command, send another prompt.
-The agent then returns the following error: _Error generating response: No output generated. Check the stream for errors._ and is blocked until the chat is cleared.
This happens with both Claude and Mistral.
You can see this in this screencast (apparently I don't have permission to upload a file):
https://streamable.com/o07vv5
Another similar error happens where, in this scenario, instead of sending another prompt, you click on the `Stop streaming` button and the chat also gets blocked. In this case, `execute-command` is stuck at `Approved - Executing` even after the cell finishes running and sending a prompt returns the same error as before: 
https://streamable.com/61q7wy

I've been thinking of a few different ways to solve this and what could be causing this and I think the simplest solution is to stop the user from sending messages while the agent is "busy" or maybe have a prompt queue. I'm still trying to understand some of the details of the implementation, but I thought I'd share my thoughts and let you know I'm working on this.

nakul-py:
>  I think the simplest solution is to stop the user from sending messages while the agent is "busy" or maybe have a prompt queue.

💯

PR #283, opened 2026-03-04
Implement agent busy state

Yahiewi:
This PR solves the issue described [here](https://github.com/jupyterlite/ai/issues/271) by adding a `_busy` and `_busyChanged` attributes to the `AgentManager` which prevent sending messages and generating a response if the agent is busy.

This screencast shows what happens when we try to send a prompt while the agent is busy:
https://github.com/user-attachments/assets/b3fb6c9d-c7a3-4c01-8e71-467335ef610c

### DIFF
--- a/src/agent.ts
+++ b/src/agent.ts
@@ -436,6 +436,20 @@ export class AgentManager {
   }
 
   /**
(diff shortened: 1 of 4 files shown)


Yahiewi:
This should be closed when PR #331 is merged.

Issue #285, opened 2026-03-06
Agent isn't aware of interrupted responses


Yahiewi:
Currently the agent isn't aware of the response it started generating if the user chooses to interrupt it using `Stop Streaming`.
History seems to be corrupted a bit when interrupting the agent:

https://github.com/user-attachments/assets/22d7c5ff-f08e-4a04-9b47-adc94f037039

This example shows a conversation that went like this:
-User: "Generate a long poem"
-Claude: *Starts generating poem titled "The Wanderer's Chronicle" then gets interrupted by the user*
-User: "What was the last prompt and what was your response to it?"
-Claude: *Generates another poem title "The Journey of Stars" first and says he has interpreted the last prompt as "Generate a long poemWhat was the last prompt and what was your response to it?"*

<img width="2190" height="356" alt="Image" src="https://github.com/user-attachments/assets/e26d689b-383e-43b8-a174-c186ce109afd" />

Sometimes the scenario is slightly different with the AI responding to only the second prompt when prompted again. However, it is always unaware of the response it started generating.

I tried this same test with **ChatGPT** and **Claude** (web versions) and they both were able to answer me correctly on the content of the response they started generating. I also tried **Copilot** which surprisingly failed at this and **Google Antigravity** which passed this test.

PR #290, opened 2026-03-16
Improve history sanitization

brichet:
Follow up #284 

Currently, the end of the message history is truncated when an unexpected message is detected (e.g. `tool-call` without result). This fix is applied when an error occurred during stream, including user abort.

This PR change this logic, in favor of including in history only the sanitized message sequence. The whole messages sequence is added to the history (after sanitization) when the response is fully generated.

It is also a workaround for #285: if the stream is stopped, the user message is not added to the history, which avoid continuous responses to an aborted message.


jtpio (approved):
Thanks!

brichet:
Test failure is not related, and hopefully will be fixed with https://github.com/jupyterlite/ai/pull/316

PR #301, opened 2026-03-24
Update to `jupyterlab-ai-commands` 0.3.1

jtpio:
Update to the latest release, which should bring improvements when editing notebooks and files.

- https://github.com/jupyter-ai-contrib/jupyterlab-ai-commands/releases/tag/v0.3.0
- https://github.com/jupyter-ai-contrib/jupyterlab-ai-commands/releases/tag/v0.3.1

Also slightly tweak the system prompt to nudge the model to get more up-to-date information about the files it edits.


brichet (approved):
Thanks @jtpio , it seems a lot more reliable.
I still see some error done by mistral (devstral), but I think it is also an issue with the LLM. Sometime when I ask something that is not fully relevant with the current notebook, it opens a lot of notebooks to see if one matches better the request.

Ideas for future PRs:
- add a command to read the content without opening the widget, to help introspecting files and opening only the relevant one. It was part of https://github.com/jupyter-ai-contrib/jupyterlab-ai-commands/pull/5, maybe we could keep only the part about reading content.
- specify to ask the user when the request is not fully understand, before starting to call tools
- I haven't see issues so far with this PR, but it could be useful to tell the agent to compare the file after the change, and correct it if it is not what was expected

jtpio:
> it opens a lot of notebooks to see if one matches better the request.

It could be because the system prompt is a bit too geared towards taking action and creating notebooks. Maybe this could indeed be tweaked.

PR #303, opened 2026-03-24
Save and restore chat

brichet:
This PR add commands and buttons to save and restore chats.

Fixe #299 

In the current state, the chats are JSON serialized and save as a file in a ~hidden~ directory.

Saving can be done manually with the toolbar button, or automatically with the small button next to the save button image. The automatic save is triggered when messages list updated, with a 3s debouncer to avoid triggering several times during streaming.

Restoring is done automatically from the chat name when restoring the session, and can also be done manually using the commands (a file is then required).

### (auto)save and auto restoration

[output.webm](https://github.com/user-attachments/assets/b1973da4-9baf-4dcf-ba64-235bb1028afd)

### manual restoration

[output.webm](https://github.com/user-attachments/assets/ae73ef6c-5212-4847-9f0e-226da7be097e)

### To do
- [x] add automatic save option
- [x] add automatic restoration when restoring the session after reload
- [ ] ~disable the command when restoration is not possible~
- [ ] fix the chat toolbar in the side panel, that may contain too many icon. We should probably use a reactive toolbar in `@jupyter/chat`
- [x] add tests


src/components/save-button.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * A split button for saving the chat, with a dropdown to toggle auto-save.
+ * When auto-save is active, the save button displays the JupyterLab
+ * toggled-on appearance (inset box-shadow all around).
+ */
+export function SaveComponent(props: ISaveButtonProps): JSX.Element {
+  const { model, translator: trans } = props;
+
+  const [autosave, setAutosave] = useState(model.autosave);
```
jtpio: Does this properly pick up the new state when the `restore()` method (setting the value of `autosave`) is called in the chat model?

src/agent.ts
```diff
@@ -471,6 +472,34 @@ export class AgentManager implements IAgentManager {
...
+      this._agentEvent.emit({
+        type: 'tool_approval_resolved',
+        data: { approvalId, approved: false }
+      });
+    }
+    this._pendingApprovals.clear();
+
+    // Convert chat messages to model messages
+    const modelMessages = messages.map(msg => {
+      const isAIMessage = msg.sender.username === 'assistant';
```
jtpio: Should this be `ai-assistant` instead? Like in a couple of places in the code base:

https://github.com/jupyterlite/ai/blob/8df7f2c5c511da05032750e0b528b88cd7f1edff/src/chat-model.ts#L274

src/components/save-button.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * Effect that update the autosave state when it is updated on the model.
+   */
+  useEffect(() => {
+    model.autosaveChanged.connect((_, value) => setAutosave(value));
+    return () => {
+      model.autosaveChanged.disconnect((_, value) => setAutosave(value));
```
jtpio: This should probably refer to the same callback function used above for `connect`.

src/agent.ts
brichet: Good catch, thanks.

jtpio (approved):
Thanks!

PR #306, opened 2026-04-02
Add support for images and pdf files

### LINKED ISSUE #292: Images are not being parsed correctly
It seems like images uploaded as attachments to the conversation are being processed as base64 and the agent isn't really able to make sense of them this way.
This also blows up the API key usage because as you can see here, a single 400kb image used ~540000 tokens which is extremely expensive for a single prompt; around 2.7$ for Claude Opus 4.6 and 0.21$ for Mistral.

Image

I'm thinking we should extend the [`_processAttachments`](https://github.com/jupyterlite/ai/blob/565b3a02e70706022dde4c2ee9394bed35ac7137/src/chat-model.ts#L645) function to correctly handle the image type.

Yahiewi:
This PR fixes #292 by fixing the way attachments are read and sent to the AI.

Main code changes:

- In the `sendMessage` function, `enhancedMessage` is now declared as [`UserContent`](https://github.com/vercel/ai/blob/07f7de8cdc3de3896133d0d3ef5f66ca592990a3/packages/provider-utils/src/types/user-model-message.ts#L22) and `attachmentContents` is now replaced by `{ textContents, binaryParts }`
- Changed `_processAttachments` return signature changed from `Promise<string[]>` to ```
Promise<{
    textContents: string[]
    binaryParts: Array<ImagePart | FilePart>}>``` and replaced `const contents` with `textContents: string[]` and `binaryParts : Array<ImagePart | FilePart>`
- Defined:
-`imageMimeTypes` : **.png**, **.jpg**, **.jpeg**, **.gif**, .**webp**, **.bmp** (which is sent as ImagePart)
-`fileMimeTypes` : **.pdf** (which is sent as FilePart)
- Defined `_readBinaryAttachment` function to take in both images and PDFs and return their base64 content. 

https://github.com/user-attachments/assets/aaae6c2f-68a0-4496-bcdb-cffad851209a


src/chat-model.ts
```diff
@@ -643,41 +652,105 @@ export class AIChatModel extends AbstractChatModel {
...
-    attachments: IAttachment[]
-  ): Promise<string[]> {
-    const contents: string[] = [];
+  private async _processAttachments(attachments: IAttachment[]): Promise<{
+    textContents: string[];
+    binaryParts: Array<ImagePart | FilePart>;
+  }> {
+    const textContents: string[] = [];
+    const binaryParts: Array<ImagePart | FilePart> = [];
+    const imageMimeTypes: Record<string, string> = {
```
brichet: Thanks @Yahiewi for working on this.

Should we use the mimetype that is defined in the `IAttachment` ? https://github.com/jupyterlab/jupyter-chat/blob/ae02ca6f6ccb5206514306f318d3a5a4a970cd45/packages/jupyter-chat/src/types.ts#L194

Also #309 included the [mimetype in Notebook attachments](https://github.com/jupyterlab/jupyter-chat/pull/383) too.

Yahiewi: I replaced the hardcoded extenstion-to-mimetype mappings with IAttachment.mimetype.
I also added a contents manager metadata fetch fallback for when there is no provided mimetype; I'm still not sure why it's not always available even after those PRs were merged.

src/chat-model.ts
```diff
@@ -651,41 +660,113 @@ export class AIChatModel extends AbstractChatModel {
...
-          const fileContent = await this._readFileAttachment(attachment);
-          if (fileContent) {
-            const fileExtension = PathExt.extname(
-              attachment.value
-            ).toLowerCase();
-            const language = fileExtension === '.ipynb' ? 'json' : '';
-            contents.push(
-              `**File: ${attachment.value}**\n\`\`\`${language}\n${fileContent}\n\`\`\``
-            );
+          let mimetype = (attachment as any).mimetype;
```
```suggestion
          let mimetype = attachment.mimetype;
```

No need for casting the type I think.

brichet (commented):
Thanks @Yahiewi for updating the PR, it works well.

Some thoughts about this change, maybe not for this PR:
- I wonder if we should not ask for confirmation before sending some files, it takes a lot of token (image or PDF can take more than 10k tokens)
- implementation details: the private functions `_processAttachments`, `_readBinaryAttachment`, `_readNotebookCells` and `_readFileAttachment` could be moved to `Private` namespace, with a `documentManager` argument. That would clarify the code, and maybe ease handling the case where `documentManager` is not available.

Yahiewi:
Thanks @brichet for the review and feedback.
I was testing token usage when sending an image and I realized because of the fact that we resend the entire history everytime a user sends a prompt, token usage can grow very quickly.
For example, in a conversation I had with Mistral, the stats looked like this:
```
[mistral-1] Token usage: input=5693, output=387
[mistral-1] Token usage: input=11782, output=477
[mistral-1] Token usage: input=17970, output=626
```
This was because I sent an image in the first prompt which kept being sent throughout the rest of the conversation. I'm not sure what's the best way to deal with this, and it seems to me that most LLMs handle their history like this too. I think we can try to cut on the token usage though:

Instead of resending attachments every time, maybe we can just save a summary which describes the image/PDF and save the memory in the cache, then when the agent needs to see it again, it can retrieve it.
This idea needs more thought if it were to be implemented but I'd love to hear what you think first.
If you think this is a good idea.I can open a separate issue and start working on something.

For now, I think we can add a modal, when sending an attachment for the first time, that warns you that attachments use a lot of tokens (and you can check a box for "Don't show this again" so it doesn't become annoying).

brichet:
> For now, I think we can add a modal, when sending an attachment for the first time, that warns you that attachments use a lot of tokens (and you can check a box for "Don't show this again" so it doesn't become annoying).

That may be useful for now when sending a large attachment.

> Instead of resending attachments every time, maybe we can just save a summary which describes the image/PDF and save the memory in the cache, then when the agent needs to see it again, it can retrieve it.

I opened https://github.com/jupyterlite/ai/issues/312 related to this issue.

src/chat-model.ts
```diff
@@ -859,61 +869,286 @@ export class AIChatModel extends AbstractChatModel {
...
+  private _currentStreamingMessage: IMessage | null = null;
+  private _nameChanged = new Signal<AIChatModel, string>(this);
+  private _contentsManager?: Contents.IManager;
+  private _autosave: boolean = false;
+  private _autosaveChanged = new Signal<AIChatModel, boolean>(this);
+  private _autosaveDebouncer: Debouncer;
+}
+
+namespace Private {
+  type IMimeBody = Partial<IRenderMime.IMimeModel> &
```
brichet: `@jupyter/chat` already export [`IMimeModelBody` ](https://github.com/jupyterlab/jupyter-chat/blob/717e101353ba91ea529478848eaf678ed90f25b4/packages/jupyter-chat/src/types.ts#L67), we could use it instead.

src/chat-model.ts
```diff
@@ -859,61 +869,286 @@ export class AIChatModel extends AbstractChatModel {
...
+    return { textContents, binaryParts };
+  }
+
+  /**
+   * Reads a binary attachment and returns its base64-encoded content.
+   * @param attachment The attachment to read
+   * @param documentManager Optional document manager for file operations
+   * @returns Base64 string or null if unable to read
+   */
+  export async function readBinaryAttachment(
```
brichet, line 1104: We should not need to export this function, as it is used only in the `Private` namespace, (as well as `readNotebookCells` and `readFileAttachment`, but it makes eslint complains.

Let's keep it for now.

brichet (commented):
Thanks @Yahiewi for updating the PR.

It seems that we should also handle errors, or even better avoid it, but I'm not sure it is possible. The screenshot below uses `devstral-latest`.

Image

Yahiewi:
> Thanks @Yahiewi for updating the PR.
> 
> It seems that we should also handle errors, or even better avoid it, but I'm not sure it is possible. The screenshot below uses `devstral-latest`.
> Image

Thanks for catching that, I hadn't tested with models that don't accept any image input. I'll add code for the error handling.

Yahiewi:
I improved the error handling in general and added a patch to avoid conversation error loops:

https://github.com/user-attachments/assets/aa0ae640-6b7a-48d6-b90a-971c67b0811a

As you can see in this demo, instead of the generic "Error generating response: No output generated. Check the stream for errors." we get a more helpful error message "Error generating response: Image input is not enabled for this model" taken directly from the API.
Now you can still talk to the AI in the same conversation since in this case we strip what's causing the error from the history, which is the attachment.
Also now when an error happens, you can see in the history that instead of having two consecutive "user" messages in the history, the error is pushed to the history as an assistant message to ensure we have the right history structure and it also makes the agent aware of what went wrong.

Yahiewi:
Also since you opened an issue #312 for this. I was thinking of creating another PR after this one is merged to maybe handle attachments better to reduce token usage as I feel that is a different issue that should have its own PR.

src/agent.ts
```diff
@@ -619,6 +621,35 @@ export class AgentManager implements IAgentManager {
...
+        // Remove attachments from history on payload rejection errors
+        if (
+          APICallError.isInstance(error) &&
+          (error.statusCode === 400 ||
+            error.statusCode === 404 ||
+            error.statusCode === 413 ||
+            error.statusCode === 415 ||
+            error.statusCode === 422)
+        ) {
+          for (const msg of this._history) {
```
brichet: The whole response history (including the user message) is now added when the response is complete (to prevent responses from overlapping). Therefore, the user message doesn't exist in the history at that stage.
Maybe the user message could be added to `responseHistory` before the try statement, and be sanitized (removing the media) in the catch statement.

Yahiewi: Thanks for the heads-up. It should be good to go now.

src/agent.ts
```diff
@@ -636,7 +632,7 @@ export class AgentManager implements IAgentManager {
             error.statusCode === 415 ||
             error.statusCode === 422)
         ) {
-          for (const msg of this._history) {
+          for (const msg of [...this._history, ...responseHistory]) {
```
brichet, line 664: The error message should only appear in the `responseHistory` object; we probably don't need to go through the entire history.
Actually, if I'm not mistaken, the user message should always be the first message of the `responseHistory`, we can probably get it with `responseHistory[0]`.

Yahiewi: This is how I did it at first too (by only looking at the last sent user message).
However in the current version, I look at the entire history because of the following case:
If a user uses a vision-capable model at first and sends it attachment(s), they then later for some reason (say they no longer have any credits) decide to switch to a non vision-capable model, upon sending a message, the conversation would enter the infinite error loop because of the attachment(s) in the history.
Here's what it looks like:

https://github.com/user-attachments/assets/3998414a-849a-46c6-be62-c20ee3825f3c

As you can see we avoid the infinite error loop because the attachment is deleted when we catch the error. However, this is still not ideal I think because you need to send a message to trigger the error and then send the prompt. I'm currently thinking about how I can remove this friction.

EDIT: I tried encapsulating the function body in a while loop and make it retry sending the message once automatically after the first error of this kind and this works fine. But I think it overcomplicates the code a bit and makes it less readable so in the end I just improved the error logging:

image

brichet: Thanks for the clarification, indeed it would break when switching the model.

Let's keep it that way for now, but we should open an issue to handle it.

brichet (approved):
Thanks @Yahiewi, it looks good.

src/agent.ts
brichet: I opened https://github.com/jupyterlite/ai/issues/321

Yahiewi:
Hey @brichet thanks for the review.
FYI, I don't think I can merge this, so please feel free to do that when you can.

Issue #288, opened 2026-03-09
Option to edit previous messages


Yahiewi:
One feature that is available in other agents that would be useful here is the option to go back to a previous message and edit it and restart the conversation from that point.
This would drop the history that comes after that message, but would also have to undo the changes made in the workspace.
Here's an example of how this is implemented in something like Copilot:

https://github.com/user-attachments/assets/a24bd22d-1dfa-4fee-a41d-542afb79836f

In this example, I prompt the agent to add some console logs, then when I modify that message I get a modal to confirm that I want to restore the history and code to that state.
We could do something similar, or perhaps even better would be to implement a branching history, so users don't lose their previous conversations or modifications when editing their messages.

I imagine the hardest part is undoing all of the different commands an agent can execute and I'm not yet sure how feasible this even is. 
We could look at this [PR](https://github.com/jupyterlab/jupyter-collaboration/pull/338/changes) for inspiration.
We might also be able to use the Jupyter checkpoint system (`.ipynb_checkpoints`).

Yahiewi:
I opened PR #310 to add the option to edit messages, this feature does not however undo the code changes made by the AI agent.
I think that needs to have a separate PR if it were to happen and is not going to be straight-forward because of the considerations in the collaborative case.

PR #310, opened 2026-04-07
Add branching conversations through an edit message option

Yahiewi:
This PR closes issue #288 by adding an option to edit previously sent user messages which creates a branch or multiple branches if that message was edited more than once.
This uses `updateMessage` from @jupyter/chat.

Code changes:

src/components/edit-message-button.tsx and src/index.ts:
- This new React component is added as a section using `createEditMessageSection: MessageFooterSection` in a new `messageFooterRegistry` (from @jupyter/chat). It has a `BranchNavigation` component which shows the branch navigation arrows on edited user messages. It also shows a counter for the branches (for example 2/3). It also preserves scroll position so when editing a message, the conversation isn't scrolled to end.

src/agent.ts:
- Added `history` getters and setters as well as a `truncateHistory(n)` function (which keeps the first `n` turns (user message and response) in the history) in the `AgentManager` class. 
EDIT: Since a `setHistory` function was also defined in PR #303, I renamed the setHistory function defined here to `restoreHistory`. Although I think it makes more sense to switch the names since the function defined in this PR is a simple setter.

src/chat-model.ts:
- Added `_branchPoints` private field in `AIChatModel` class of type `Map<string,{ branches: AIChatModel.IBranchSnapshot[]; currentIndex: number }` which maps to every edited message `messageId` an array of branches and the currently active branch index.
- `clearMessages` method now also clears the branch points.
- Added several functions: `editMessage`, `switchBranch`, `updateMessage` (which is called by @jupyter/chat's native edit UI and basically calls editMessage), `getBranchInfo`, a new private `_snapshotFrom` method and refactored out a `_createErrorMessage` method that's used in `editMessage` and `sendMessage`.
- In the `AIChatModel` namespace, added an `IBranchSnapshot` interface.

Demo:

https://github.com/user-attachments/assets/7197520e-7b7a-4e6f-9873-978cd94ac09e


Yahiewi:
I'm thinking about how we can undo the code changes when editing a message, or maybe give the user the option to undo the the code changes. However, I'm still not quite sure how to do this or if this is doable. The main challenge I think is implementing this in a collaborative context:
-If a user undoes his changes by editing the message, should this also undo the changes for his collaborators? What happens if a user is editing a section of the code which is generated by the AI and has been undone by his collaborator? I haven't worked on collaborative issues in Jupyter, so I'm not sure how feasible this is.

PR #316, opened 2026-04-10
Fix timeout in tests on save and restore chat

### LINKED ISSUE #315: Fix flaky tests on save and restore chat
## Description

The tests on saving and restoring chats, introduced in #303, seem to be flaky, we should fix them.

brichet:
Fix #315


brichet:
Merging it, it's only about adding timeout to tests.

PR #320, opened 2026-04-13
Specify chat file format (backup) when fetching content

brichet:
Some contents manager may use 'base64' as default format, which ends up as an error when trying to parse the JSON content.

This PR specify the type and format of the content when fetching it.


jtpio (approved):
Thanks!

Issue #321, opened 2026-04-14
Handling attachments capabilities of model


brichet:
### Problem

#306 handles files and images attachments to messages.
These attachments are currently added to the context, in the history.

However, some models doesn't have the capability to read them, which results to an error. Therefore, the attachment has to be removed from the user message when the error occurs.
When switching from a model (with vision capability) to another (without vision capability), the error can come from a previous message. This is currently handled by going through the entire history to find attachments and remove them (see https://github.com/jupyterlite/ai/pull/306#discussion_r3080471601).

### Proposed Solution

After #262, we have a way to fill capabilities per model 
https://github.com/jupyterlite/ai/blob/0f8ccdcad878b5b9be17eae3d29cfa0c9f93a06d/src/tokens.ts#L207

Currently, this is used to get the context window available (https://github.com/jupyterlite/ai/blob/main/src/providers/generated-context-windows.ts) per model, computed with https://github.com/jupyterlite/ai/blob/main/scripts/sync-model-context-windows.mjs.

The information comes from https://models.dev/, which also provides an input field, that could probably be useful to build the list of input capabilities per model.

Using that, we could rebuild the history from the chat message when switching of model, removing the attachments if not supported.

PR #323, opened 2026-04-16
Improve attachment input handling and history

Yahiewi:
This addresses issue #321 by using the database from https://models.dev/ to map which models accept images as input.
The main changes now are:
- When using a non vision capable model, the "attach button" is no longer visible. (You can still drag and drop an image and get the error).
- When switching from a vision capable model (say Claude Opus 4.6) to a non capable one (devstral-latest for example), the attachments are automatically stripped from the chat without needing the user to send a message first to catch the error which used to be the [case](https://github.com/jupyterlite/ai/pull/306#discussion_r3080471601).

Code changes:

The info is stored under `generated-model-info.ts` which is generated using `jlpm sync:model-info`.
The script was previously called `scripts/sync-model-context-window.mjs` but it was renamed and extended to fetch both the context window and if or not models can take images as input.
A new boolean `modelSupportsImages` function was added to `model-info.ts` which returns true in the absence of `modelInfo`; since it's not possible to know beforehand when we have no modelInfo (for local models for example), I think the safe option is to assume that it's vision capable.
Added a new `_lastModelKey` attribute to the AgentManager class which helps detect if the model has changed in the `_prepareAgentConfig` function; if it has changed to a non vision capable model, the new `_stripAttachments` function is used to remove attachments from history.


src/agent.ts
```diff
@@ -661,19 +664,10 @@ export class AgentManager implements IAgentManager {
...
-                const textContent = msg.content
-                  .filter(p => p.type === 'text')
-                  .map(p => (p as { text: string }).text)
-                  .join('\n');
-                msg.content =
-                  textContent || '_Attachment removed due to error_';
-              }
-            }
-          }
+          this._stripAttachments(
```
brichet, line 645: Couldn't we avoid the error ?
I mean not adding the binary data if the model doesn't support it, but either adding an information that a file has been attached but the content is not provided.

src/agent.ts
```diff
@@ -793,6 +808,20 @@ export class AgentManager implements IAgentManager {
       this._providerRegistry
     );
 
+    // Strip attachments from history if the new model doesn't support them.
+    const currentModelKey = activeProviderConfig
+      ? `${activeProviderConfig.provider}:${activeProviderConfig.model}`
+      : undefined;
+    if (currentModelKey && currentModelKey !== this._lastModelKey) {
+      this._lastModelKey = currentModelKey;
+      if (!modelSupportsImages(activeProviderConfig, this._providerRegistry)) {
```
brichet: Would there be a way to restore the attachment if the model support it ?

src/index.ts
```diff
@@ -557,6 +560,22 @@ const plugin: JupyterFrontEndPlugin<IChatTracker> = {
         );
       }
 
+      // Show/hide attach button based on whether the active model supports images.
+      function updateAttachButton() {
+        const providerConfig = settingsModel.getProvider(
+          model.agentManager.activeProvider
+        );
+        if (modelSupportsImages(providerConfig, providerRegistry)) {
```
brichet: It might be too restrictive, the model could still support attaching text file.

brichet (commented):
Thanks @Yahiewi for working on this.

I did not go through the code yet, but I have some general comments about the PR.

1. when adding a message to the history, we should check if the model supports images, and add or not the binary data. That would avoid throwing an error.
2. we should probably regenerate the history from the chat messages when the provider is updated (instead of just removing the attachments if it is not supported, see [comment](https://github.com/jupyterlite/ai/pull/323#discussion_r3101494529)).
    With the suggestion above, the history would be sanitized by default when regenerated.
3. I wonder if we could update the attachment preview if the file is not supported Image.
    And maybe we could also update the tooltip (`attachment not supported by the model`), but I'm not sure if this is possible with the current implementation of `jupyter-chat`.
    The input model has an `attachmentsChanged` [signal](https://github.com/jupyterlab/jupyter-chat/blob/9f7ee280142bf7de497931f84c06c956c7e3cbc0/packages/jupyter-chat/src/input-model.ts#L128) that could be helpful for that.

Yahiewi:
> Thanks @Yahiewi for working on this.
> 
> I did not go through the code yet, but I have some general comments about the PR.
> 
>     1. when adding a message to the history, we should check if the model supports images, and add or not the binary data. That would avoid throwing an error.
> 
>     2. we should probably regenerate the history from the chat messages when the provider is updated (instead of just removing the attachments if it is not supported, see [comment](https://github.com/jupyterlite/ai/pull/323#discussion_r3101494529)).
>        With the suggestion above, the history would be sanitized by default when regenerated.
> 
>     3. I wonder if we could update the attachment preview if the file is not supported Image.
>        And maybe we could also update the tooltip (`attachment not supported by the model`), but I'm not sure if this is possible with the current implementation of `jupyter-chat`.
>        The input model has an `attachmentsChanged` [signal](https://github.com/jupyterlab/jupyter-chat/blob/9f7ee280142bf7de497931f84c06c956c7e3cbc0/packages/jupyter-chat/src/input-model.ts#L128) that could be helpful for that.

Thanks @brichet for the review. I made a few changes in the last commit to address the points you raised:
1. We now avoid the error and don't need it to be raised for the attachment to be removed. In the `sendMessage` function we now check if the model supports images and if it does, keep the (image and PDF) attachments, and if it doesn't, remove them and add some fallback text. Text files are not concerned with this.
2. Attachments can now be restored to the history in the case of switching from a model that doesn't support images to one that does with the new `_onModelChanged` which detects the change using the new private field `currentModelKey` and calls `_rebuildHistory` which rebuilds the history of the conversation depending on the model's capabilites.
3. I'm not sure we can do this without making a PR for jupyter/chat. I think we can leave the UI as is right now and maybe work on this in another PR.

Here's a quick demo:

https://github.com/user-attachments/assets/5a35bc75-badf-4ea2-b468-ae4814114455

src/chat-model.ts
```diff
@@ -511,6 +536,69 @@ export class AIChatModel extends AbstractChatModel {
...
+          role: 'user',
+          content:
+            binaryParts.length > 0
+              ? [{ type: 'text', text: textPart }, ...binaryParts]
+              : textPart
+        } as ModelMessage);
+      } else {
+        modelMessages.push({
+          role: isAI ? 'assistant' : 'user',
+          content: msg.body
```
brichet, line 660: We should handle empty `body` (MIME renderer messages) by either removing them or adding a body that describes the content that should be there.

src/agent.ts
```diff
@@ -517,6 +517,26 @@ export class AgentManager implements IAgentManager {
     this._history = Private.sanitizeModelMessages(modelMessages);
   }
 
+  /**
+   * Sets the history from already-processed model messages.
+   * Allows including binary content for vision-capable models.
+   * @param messages Pre-built model messages
+   */
+  setPreprocessedHistory(messages: ModelMessage[]): void {
```
brichet: This method is used to do the same thing as `setHistory` method: restoring agent history from chat messages.
We should probably merge them into a unique workflow, to avoid duplication, keeping the workflow in this PR, as it handles attachments.
I would remove the previous `setHistory()` method, rename this one `setHistory()`, and use the new `_rebuildHistory` (in chat model) when restoring the chat.

brichet (commented):
Thanks @Yahiewi for updating it.

There is an issue (not directly because of this PR, it also happens with restored chat), depending on the provider.
The tool calls cannot be restored in history from the chat messages, otherwise it would require some hand made tool call history, that we certainly want to avoid.
The tool call messages are rendermime messages without body content (as other rendermime message by the way). These messages appear as empty `assistant` message in the history, which seems to break 'Claude' model (at least), but not with `Devstral`:
`Error generating response: messages: text content blocks must be non-empty`.

We should probably handle it, by adding a text content about tool used (or rendermime content), or at least remove the empty messages.

Yahiewi:
Thanks for the rereview @brichet 
Here are the changes I made:
- In `_rebuildHistory` we now skip messages with empty bodies (to avoid the Claude error).
- I removed the old `setHistory` and renamed `setPreprocessedHistory` to `setHistory`, the `restore` function now uses `_rebuildHistory`. This has several advantages: it avoids the Claude error you mentioned, and it also restores attachments when rebuilding the history.

Here's a demo showcasing the second case (I also tested for the tool call case and we no longer get the empty body error):

https://github.com/user-attachments/assets/5d484a0d-ff51-49dd-9a11-7e5785b6d76a

Yahiewi:
Hmm the `Check Release` job seems to be failing for some reason. 
Copilot says rerunning the job may make it pass but I don't think I have permission to do that.

nakul-py:
> Hmm the `Check Release` job seems to be failing for some reason.

Yep sometimes `Check Release` behaves like this.

Yahiewi:
Okay looks like it's passing now.

src/chat-model.ts
```diff
@@ -341,9 +356,19 @@ export class AIChatModel extends AbstractChatModel {
             '\n\n--- Attached Files ---\n' + textContents.join('\n\n');
         }
 
-        if (binaryParts.length > 0) {
+        const providerConfig = this._settingsModel.getProvider(
+          this._agentManager.activeProvider
+        );
+        if (
+          binaryParts.length > 0 &&
+          modelSupportsImages(providerConfig, this._providerRegistry)
```
brichet: This prevent attaching document or audio if images are not supported.

src/chat-model.ts
```diff
@@ -559,6 +584,70 @@ export class AIChatModel extends AbstractChatModel {
...
+    );
+    const supportsImages = modelSupportsImages(
+      providerConfig,
+      this._providerRegistry
+    );
+
+    const modelMessages: ModelMessage[] = [];
+    for (const msg of this.messages) {
+      const isAI = msg.sender.username === 'ai-assistant';
+      if (!isAI && msg.attachments?.length && supportsImages) {
```
brichet: Same as above. 
Maybe it should be handled in `processAttachments` instead, where we process images and others types specifically.

src/chat-model.ts
```diff
@@ -341,9 +356,19 @@ export class AIChatModel extends AbstractChatModel {
...
+        );
+        if (
+          binaryParts.length > 0 &&
+          modelSupportsImages(providerConfig, this._providerRegistry)
+        ) {
           enhancedMessage = [{ type: 'text', text: textPart }, ...binaryParts];
         } else {
+          if (binaryParts.length > 0) {
+            textPart +=
+              '\n[Binary attachments omitted: this model does not support image inputs.]';
```
brichet: Should we add the file path to give some more information to the model ?
This could also be in `processAttachments`, to include it when using `_rebuildHistory` too.

src/agent.ts
```diff
@@ -506,16 +504,7 @@ export class AgentManager implements IAgentManager {
     }
     this._pendingApprovals.clear();
```
brichet:
```suggestion
this.stopStreaming('Chat history changed');

```

This is a leftover form a previous PR (not because of this one), but the pending approvals should already be resolved when using `stopStreaming`.

It can be included in another PR, as it is not due to this change.

brichet (commented):
Thanks @Yahiewi for the update.

src/agent.ts
brichet: The suggestion is not working, but it was about removing all the part about approving and clearing the pending approvals.

Yahiewi:
This last commit brings a few changes:
-I addressed all you review comments and made the requested changes.
-Since you mentioned added support for audio, I wanted to see what it would take to add support for that because I hadn't considered it yet. In the [models.dev](https://models.dev/) database, we can check if a model supports audio input or not. So just like I added the `supportsImages` field for the providerModelInfo, I added `supportsAudio`.
Claude and Mistral both don't support audio input, so I created a free Gemini API key and tested using the `gemini-flash-latest` which does.
In this demo you can see Gemini accurately describing a 3 second synth melody and transcribing a 5s audio sample:

https://github.com/user-attachments/assets/652182fd-19da-481f-baa0-63cb387a358d

Also, since some models like `gemini-flash-latest` support video input too, I think it should also be possible to add this feature, but I think that is better left to another PR, as there is the issue of these files being much larger than other attachments and I think it's better to look for a way to handle that before adding this feature.

brichet:
The code is duplicated between `sendMessage()` and `_rebuildHistory()`.

Why not handling it in `processAttachments` instead ?
It could check for the capacities of the model, and return either the binary content or a text content with `\n[Attachments omitted (not supported by this model): ${omittedNames.join(', ')}.]` if not.

Yahiewi:
Thanks for catching that.
I refactored that logic into `processAttachments` now to handle the potential filtering of attachments in the main loop. 
I also changed its signature accordingly: by adding `body` and `supportsX` booleans as arguments and having it return `Promise<UserContent>` directly.
I also simplified a few things.
I tested again with the different scenarios and it seems to be working fine.

brichet (approved):
Thanks @Yahiewi, LGTM.
🚀

Yahiewi:
Thanks @brichet, I don't seem to have permission to merge, so please feel free to do it when you can.

PR #324, opened 2026-04-16
Remove INotebookTracker from IChatTracker plugin dependencies

brichet:
Similar to https://github.com/jupyterlite/ai/pull/258 for the chat tracker.

It allows to use the `IChatTracker` in area like the launcher, which is awaited by the notebook tracker plugin (and therefore creates circular dependencies).

The active cell manager has now its own plugin.


src/index.ts
```diff
@@ -416,17 +440,6 @@ const plugin: JupyterFrontEndPlugin<IChatTracker> = {
       }
     };
 
-    // Create ActiveCellManager if notebook tracker is available, and add it to the
-    // model registry.
-    let activeCellManager: ActiveCellManager | undefined;
-    if (notebookTracker) {
```
jtpio, line 422: Looks like `INotebookTracker` is still declared in the list of `optional` dependencies for this plugin?

brichet: Oups, sorry for that, I fixed it.

src/index.ts
```diff
@@ -357,6 +357,30 @@ const chatModelHandler: JupyterFrontEndPlugin<IChatModelHandler> = {
...
+  description: 'Add the active cell manager to the model handler',
+  autoStart: true,
+  requires: [IChatModelHandler, INotebookTracker],
+  activate: (
+    app: JupyterFrontEnd,
+    modelHandler: IChatModelHandler,
+    notebookTracker: INotebookTracker
+  ) => {
+    let activeCellManager: ActiveCellManager | undefined;
+    if (notebookTracker) {
```
jtpio: Also do we still need this check if `INotebookTracker` is required?

brichet: :+1: I definitely pushed this PR too quickly...

Thanks for raising it.

jtpio (approved):
Thanks!

PR #325, opened 2026-04-16
Await for chat messages to be cleared before setting history

brichet:
Follow up #303, the `clearMessages` method was not properly awaited, and the newly added history was cleared as well.

My bad, these changes have been suggested [here](https://github.com/jupyterlite/ai/pull/290#discussion_r2947529030), but back then, the clear messages method wasn't used that way...

### DIFF
--- a/src/chat-model.ts
+++ b/src/chat-model.ts
@@ -249,7 +249,7 @@ export class AIChatModel extends AbstractChatModel {
   clearMessages = async (): Promise<void> => {
     this.messagesDeleted(0, this.messages.length);
     this._toolContexts.clear();
-    this._agentManager.clearHistory();
+    await this._agentManager.clearHistory();
   };
 
   /**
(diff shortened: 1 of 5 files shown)


src/chat-model.ts
```diff
@@ -246,10 +246,10 @@ export class AIChatModel extends AbstractChatModel {
   /**
    * Clears all messages from the chat and resets conversation state.
    */
-  clearMessages = (): void => {
+  clearMessages = async (): Promise<void> => {
```
jtpio, line 249: Should the public facing method also be updated here?

https://github.com/jupyterlite/ai/blob/72f6e60d413de7944e7cdcdaa474675eac33a6f4/src/chat-model.ts#L1432

brichet: Good catch, I updated it and also the one in `clearButton`.
https://github.com/jupyterlite/ai/blob/72f6e60d413de7944e7cdcdaa474675eac33a6f4/src/components/clear-button.tsx#L19

It also properly awaits each call of `clearMessage`.

jtpio (approved):
Thanks!

PR #327, opened 2026-04-17
Chat model title

brichet:
This PR add an optional title to the chat model.
When set, this title appears as an overlay on the chat name (tab for main area widget or toolbar for side panel widget). It is also saved in the backup file, and restored.

It also adds methods to generate a title:
- `textResponse` in agent, that creates a transient model and request a text from a sequence of messages.
- `requestTitle` in chat model, that create a simple prompt from messages to request a title to the chat, using the `textResponse` method.

It can be tested from the console (needs a chat with content to be relevant):

```js
model = jupyterapp.pluginRegistry._plugins.get('@jupyterlite/ai:plugin').service.currentWidget.model;
model.title = await model.requestTitle()
```
It will generate a title related to the chat history, and set it as title of the chat model.

### DIFF
--- a/src/agent.ts
+++ b/src/agent.ts
@@ -706,6 +706,7 @@ export class AgentManager implements IAgentManager {
         model,
         messages
       });
+      this._updateTokenUsage(result.totalUsage, result.totalUsage.inputTokens);
       return result.text;
     } catch (e) {
       throw `Error while getting the topic of the chat\n${e}`;
--- a/src/agent.ts
+++ b/src/agent.ts
@@ -4,6 +4,7 @@ import { IRenderMimeRegistry } from '@jupyterlab/rendermime';
 import { PromiseDelegate } from '@lumino/coreutils';
 import { ISignal, Signal } from '@lumino/signaling';
 import {
+  generateText,
   ToolLoopAgent,
   type ModelMessage,
   type LanguageModel,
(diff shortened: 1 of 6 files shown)


src/agent.ts
```diff
@@ -694,6 +695,23 @@ export class AgentManager implements IAgentManager {
...
+   * @param messages - the messages sequence to send to the model.
+   */
+  async textResponse(messages: ModelMessage[]): Promise<string> {
+    try {
+      const model = await this._createModel();
+      const result = await generateText({
+        model,
+        messages
+      });
+      return result.text;
```
jtpio, line 710: Should we also update the token usage here too?

Something like the following maybe:


```suggestion
      this._updateTokenUsage(result.totalUsage, result.totalUsage.inputTokens);
      return result.text;
```

src/chat-model.ts
```diff
@@ -136,6 +136,31 @@ export class AIChatModel extends AbstractChatModel {
...
+
+  /**
+   * The title of the chat.
+   */
+  get title(): string | null {
+    return this._title;
+  }
+  set title(value: string | null) {
+    this._title = value;
+    this._titleChanged.emit(this._title);
```
jtpio, line 157: Should this also trigger the autosave?

Something like the following?


```suggestion
    this._titleChanged.emit(this._title);
    if (this.autosave) {
      this._autosaveDebouncer.invoke();
    }
```

jtpio (approved):
Thanks @brichet! 

Left a couple of minor comments. Should be good to go otherwise 👍

src/agent.ts
brichet: Good catch, thanks

brichet:
Thanks for the review and suggestions, I added a commit.

jtpio (approved):
Thanks!

PR #328, opened 2026-04-17
Use grouped tool calls component

brichet:
1. The first intention of this PR is to replace the [tool call](https://github.com/brichet/jupyter-chat-components/blob/8c60e70b83fe49622fe85db99f5ea0109f950f52/src/components/tool-call.tsx#L122) component by the [grouped tool calls](https://github.com/brichet/jupyter-chat-components/blob/8c60e70b83fe49622fe85db99f5ea0109f950f52/src/components/grouped-tool-calls.tsx#L681) component.

    There is probably no need to maintain both tool call components in the long run, so we could use the same as `jupyter-ai`.

    [output.webm](https://github.com/user-attachments/assets/24f10eb0-9e0f-4670-a956-fe1fa6c64809)

2. It also removes the `approvalID` that is shared during the awaiting approval process.

    In the current implementation, there can be only one approval ID per tool call ID, so we don't really need it outside of the `_handleApprovalRequest` of the agent.
    This change can easily be reverted if not expected (could we have several approval ID for one tool call ? )

### DIFF
--- a/demo/uv.lock
+++ b/demo/uv.lock
@@ -743,11 +743,11 @@ wheels = [
 
 [[package]]
 name = "jupyter-chat-components"
-version = "0.4.0"
+version = "0.4.1"
 source = { registry = "https://pypi.org/simple" }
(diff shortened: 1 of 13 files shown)


brichet:
I updated the demo to use the latest changes in `jupyter-chat-components` https://github.com/brichet/jupyter-chat-components/releases/tag/v0.4.1.

jtpio (approved):
Thanks!

Testing the tool rejection path on the RTD preview gives the following error, although it's very likely a pre-existing issue not related to this change:

Image

brichet:
Thanks for the review.

> Testing the tool rejection path on the RTD preview gives the following error, although it's very likely a pre-existing issue not related to this change:

I can't reproduce it, but yes, it should be related to this change.

PR #329, opened 2026-04-19
Add Open-or-Reveal Chat Command with Prompt Prefill Support

MUFFANUJ:
closes #307 

This PR adds a public command flow for opening or revealing AI chat and pre-filling message input from extensions. It introduces a dedicated open-or-reveal command, supports optional `input`/`focus`/`autoSend` behavior for chat opening.


src/index.ts
```diff
@@ -801,11 +859,18 @@ function registerCommands(
           return false;
         }
 
+        const shouldFocus = args.focus !== false;
```
jtpio: Wondering if the focus should be applied only when passed explicitly?



```suggestion
        const shouldFocus = args.focus === true;
```

src/index.ts
```diff
@@ -801,11 +859,18 @@ function registerCommands(
...
-          openInMain(model);
+          widget = openInMain(model);
         } else {
-          chatPanel.open({ model });
+          widget = chatPanel.open({ model });
+        }
+        if (shouldFocus) {
+          focusOnChat(area, widget);
         }
+        applyInputArgs(model, args);
```
jtpio: And pass it down explicitly here?

```suggestion
        applyInputArgs(model, { ...args, focus: shouldFocus });
```

jtpio (commented):
Thanks @MUFFANUJ for working on this 👍 

Looks good overall, left a few minor comments.

src/index.ts
```diff
@@ -767,6 +767,64 @@ function registerCommands(
         model.nameChanged.disconnect(saveTracker);
         model.agentManager.activeProviderChanged.disconnect(saveTracker);
       });
+
+      return widget;
+    };
+
+    const focusOnChat = (
```
brichet, line 782: Should we use the `focus` method ?

```ts
widget?.model.input.focus()
```

src/index.ts
```diff
@@ -767,6 +767,64 @@ function registerCommands(
...
+      const shouldFocus = args.focus !== false;
+
+      if (input !== undefined) {
+        model.input.value = input;
+      }
+      if (autoSend && input !== undefined) {
+        model.input.send(model.input.value);
+      }
+      if (shouldFocus) {
+        model.input.focus();
```
brichet, line 805: Is this required ?
It seems the chat is already focused (if set in args) before each call of `applyInputArgs`

src/index.ts
```diff
@@ -767,6 +767,64 @@ function registerCommands(
...
+      provider?: string
+    ): ChatWidget | MainAreaChat | undefined => {
+      if (!name && !provider) {
+        return;
+      }
+      return tracker.find(widget => {
+        const model = widget.model as AIChatModel;
+        return (
+          (!name || widget.model.name === name) &&
+          (!provider || model.agentManager.activeProvider === provider)
```
brichet, line 820: I wonder if we can simplify the condition.

```suggestion
          widget.model.name === name ||
          model.agentManager.activeProvider === provider
```

src/index.ts
```diff
@@ -824,6 +889,131 @@ function registerCommands(
...
+            : settingsModel.getDefaultProvider();
+          existingWidget = findChatWidget(undefined, providerConfig?.id);
+        }
+
+        // If the side chat model is loaded but not currently displayed, reveal it first.
+        if (!existingWidget && name) {
+          const loadedModel = chatPanel.getLoadedModel(name);
+          if (loadedModel) {
+            chatPanel.open({ model: loadedModel });
+            existingWidget = findChatWidget(name, provider);
```
brichet: `open` method should return the widget.


```suggestion
            existingWidget = chatPanel.open({ model: loadedModel });
```

src/index.ts
```diff
@@ -903,6 +1093,13 @@ function registerCommands(
 
         if (area === 'main') {
           openInMain(model);
+
+          if (previousWidget instanceof ChatWidget) {
```
brichet, line 1111: Is it related to the PR or a bug fix ?

brichet (commented):
Thanks @MUFFANUJ for working on this.
I have some commants/questions below.

src/index.ts
MUFFANUJ: I kept `focusOnChat` because it activates the shell area/tab (main tab or side panel). `widget?.model.input.focus()` only emits a focus signal, and it does not activate/reveal shell panels.

src/index.ts
MUFFANUJ: Yes, I think it is still needed. Activating the panel does not always guarantee the text cursor is inside the input. This call makes the focus behavior reliable when focus is requested.

src/index.ts
MUFFANUJ: I think OR logic can match a wrong chat when only one side matches.

src/index.ts
MUFFANUJ: It is related to this PR flow. `open-or-reveal` when moving side -> main, we must clean up the old side-loaded model; otherwise, we can keep stale entries / duplicate behavior. Will add a comment there to make it clearer.

src/index.ts
brichet: Right, maybe we should fix it (in a follow PR), by listening to the signal when the model is created.

src/index.ts
brichet: For consistency, the same change should be applied to the new command `open-or-reveal-chat`, as well as the change in the comment below.

brichet: Otherwise the PR looks good to me.

src/index.ts
MUFFANUJ: Sure thing! will open a PR for that.

brichet (approved):
Thanks @MUFFANUJ, looks good

PR #302, opened 2026-03-24
Implement message queue component

Yahiewi:
This branch is checked out from PR #276 since I mostly followed the `tool-call` component example to implement the message queue.
This PR implements the `message-queue` React component defined in this [jupyter-chat-components PR](https://github.com/brichet/jupyter-chat-components/pull/9).

PR #331, opened 2026-04-23
Implement queue component

Yahiewi:
This PR reimplements the message queue from PR #302 this time checked out from the main branch. This also bumps jupyter-chat-components to 0.4.0 in package.json and pyproject.toml
Here are the main code changes:
chat-model.ts:

- Added `removeQueuedMessage` callback to `AIChatModel` class which allows users to remove messages from the queue.
- Added the following private fields to  `AIChatModel`:

-  `_messageQueue: Private.IQueuedItem[] = []` with `IQueuedItem {
    id: string;
    body: string;
    _originalMsg: IMessageContent;
  }` which is an array which will contain our queued messages.
- `_isBusy : boolean = false` which is used to know whether to send a message or queue it.
- `_queueMessageId: string | null = null` which is an id used for identifiying our queue.

- Updated `sendMessage` to intercept sent message while busy and push them to the queue.
- Added `_updateQueueUI` method which controls the UI of our queue and ensures we only have one queue at a time max, it also ensures our queue stays at the bottom of the conversation.
- Added `_drainQueue` which drains the queue by pushing queued messages to the chat when agent is no longer busy.

index.ts:
- We now show the send button even if the agent is busy (same in main-area-chat.ts).
- Registered the `removeQueuedMessage` callback to handle removing items from the queue when users click the 'x'.

Here's a quick demo:

https://github.com/user-attachments/assets/996fafb5-7706-4ab8-ac76-f06edec6f0f5


Yahiewi:
There is a small UI bug when adding attachments to a queued message because the MessageQueue component from jupyter-chat-components wasn't built with that in mind. Currently, sending attachments in a queued message works fine but the attachment just doesn't appear while the message is queued.
I think it's better if this gets merged with jupyter-chat-components 0.4.0 and I'm already planning on improving the queue component by making the messages draggable (to change their queue position), I'll also add a UI for attachments in that same PR.

Yahiewi:
Great, the CI is finally passing.

src/widgets/main-area-chat.ts
```diff
@@ -103,7 +103,7 @@ export class MainAreaChat extends MainAreaWidget<ChatWidget> {
     );
 
     if (aiWriting) {
-      this.content.inputToolbarRegistry?.hide('send');
+      this.content.inputToolbarRegistry?.show('send');
```
brichet: We can probably remove this, as well as the same in `else` statement.

src/index.ts
```diff
@@ -585,7 +585,7 @@ const plugin: JupyterFrontEndPlugin<IChatTracker> = {
         );
 
         if (aiWriting) {
-          widget.inputToolbarRegistry?.hide('send');
+          widget.inputToolbarRegistry?.show('send');
```
brichet: Same comment as above.

src/chat-model.ts
```diff
@@ -361,8 +376,143 @@ export class AIChatModel extends AbstractChatModel {
...
+          msg => msg.id === this._queueMessageId
+        );
+        if (existingMsg) {
+          const idx = this.messages.indexOf(existingMsg);
+          if (idx !== -1) {
+            this.messagesDeleted(idx, 1);
+          }
+        }
+        this._queueMessageId = null;
+      }
```
brichet, line 465: This could be done without the `if (this._messageQueue.length === 0)` statement, since it is done anyway later.

And maybe we could add a private function for it, it would be useful in `_drainQueue()` too.

src/chat-model.ts
```diff
@@ -361,8 +376,143 @@ export class AIChatModel extends AbstractChatModel {
...
+      },
+      metadata: {
+        messages: this._messageQueue.map(m => ({
+          id: m.id,
+          body: m.body,
+          attachments: m._originalMsg.attachments
+        })),
+        targetId: this.name
+      } as any
+    };
```
```suggestion
      } 
    } as IMimeModelBody;
```

src/chat-model.ts
```diff
@@ -361,8 +376,143 @@ export class AIChatModel extends AbstractChatModel {
...
+      return;
     }
+
+    // Dequeue and push to chat
+    const next = this._messageQueue.shift()!;
+    next._originalMsg.time = Date.now() / 1000;
+    this.messageAdded(next._originalMsg!);
+    this._updateQueueUI();
+
+    try {
```
brichet: I wonder if we could not partially reuse the logic in `sendMessage` instead of duplicating the code.

src/chat-model.ts
```diff
@@ -612,6 +762,7 @@ export class AIChatModel extends AbstractChatModel {
     this.messageAdded(aiMessage);
     this._currentStreamingMessage =
       this.messages.find(message => message.id === aiMessage.id) ?? null;
+    this._updateQueueUI();
```
brichet: Is this needed ?
Same question for `_computeShouldAutoRenderMimeBundles` and `_handleErrorEvent`.

src/chat-model.ts
```diff
@@ -945,21 +1098,32 @@ export class AIChatModel extends AbstractChatModel {
...
 
 namespace Private {
-  type IDisplayOutput =
+  export interface IQueuedItem {
+    id: string;
+    body: string;
+    _originalMsg: IMessageContent;
+  }
+
+  export type IDisplayOutput =
```
brichet: We probably don't need to export all these constants.

brichet (commented):
Thanks @Yahiewi for working on this.
I have some comments below.

src/chat-model.ts
```diff
@@ -252,7 +252,8 @@ export class AIChatModel extends AbstractChatModel {
       stopStreaming: () => this.stopStreaming(),
       clearMessages: () => this.clearMessages(),
       agentManager: this._agentManager,
-      addSystemMessage: (body: string) => this.addSystemMessage(body)
+      addSystemMessage: (body: string) => this.addSystemMessage(body),
+      removeQueuedMessage: (id: string) => this.removeQueuedMessage(id)
```
brichet: Is the `removeQueuedMessage` in the context necessary or a left over ?

src/chat-model.ts
brichet: One idea could be to have the `sendMessage()` and the `_drainQueue()` functions calling a 
```ts
private _sendMessage(userMessage: IMessageContent)
```

src/chat-model.ts
brichet: Looks like GitHub has some issues when commenting several lines.
The comment was about 
```ts
      if (this._queueMessageId) {
        const existingMsg = this.messages.find(
          msg => msg.id === this._queueMessageId
        );
        if (existingMsg) {
          const idx = this.messages.indexOf(existingMsg);
          if (idx !== -1) {
            this.messagesDeleted(idx, 1);
          }
        }
        this._queueMessageId = null;
      }
```

src/chat-model.ts
Yahiewi: Thanks for the catch, I refactored the code into a new `_processMessage()` function used by `sendMessage` and `_drainQueue`. I chose `processMessage` over `_sendMessage` because I felt the former is more accurate, and it would be confusing to have `sendMessage` and `_sendMessage`.

Yahiewi:
Thanks @brichet for the review and the suggestions. For your specific code reviews, I replied/reacted to them directly and I addressed all of them. This comment is more about the other changes. The last few commits have a few important changes and bug fixes, not all of them necessarily related to the queue component. :

-Issue #338 is now fixed by using the same chat model when switching views, instead of creating a new one and copying the attributes of the previous model. This makes switching views much more seamless, as it doesn't start a new conversation but keeps the same one, even in the presence of queues or active tool calls. You can see this in the demo below.

-I updated `clearMessages()` to  call `stopStreaming() `and reset `_isBusy = false` and `_messageQueue = []` so that clearing the chat works properly now and we don't have stuck message queues.

-Added a filter to skip queued messages in `_serializeModel()` to avoid having them saved in the `.chat` files.


https://github.com/user-attachments/assets/72360b29-00be-4c7a-816a-c1da675dec17

There are still a couple of (non-breaking) bugs with the message queue that I'm working on in PR #334 and [PR #20](https://github.com/jupyter-ai-contrib/jupyter-chat-components/pull/20) from the jupyter-chat-components repo, which will also have new features for the queue component.

src/chat-model.ts
```diff
@@ -405,6 +431,106 @@ export class AIChatModel extends AbstractChatModel implements IAIChatModel {
...
+        }
+        this._queueMessageId = null;
+      }
+      return;
+    }
+
+    // Dequeue and push to chat
+    const next = this._messageQueue.shift()!;
+    next._originalMsg.time = Date.now() / 1000;
+    this.messageAdded(next._originalMsg!);
```
```suggestion
    this.messageAdded(next._originalMsg);
```
The exclamation mark should not be necessary.

src/chat-model.ts
```diff
@@ -621,7 +754,7 @@ export class AIChatModel extends AbstractChatModel implements IAIChatModel {
    * For vision-capable models, re-reads binary attachments from disk.
    * For text-only models, uses message text only.
    */
-  private async _rebuildHistory(): Promise<void> {
+  rebuildHistory = async (): Promise<void> => {
```
brichet: Do we need to expose the `rebuildHistory` method ?

src/chat-model.ts
```diff
@@ -1045,6 +1181,30 @@ export class AIChatModel extends AbstractChatModel implements IAIChatModel {
 
   // Private fields
   private _settingsModel: IAISettingsModel;
+  /**
+   * The current message queue
+   */
+  get messageQueue(): Private.IQueuedItem[] {
```
brichet, line 1191: We should move `getMessageQueue` and `isBusy` getters/setters above the private fields.

src/index.ts
```diff
@@ -1083,61 +1094,25 @@ function registerCommands(
...
-
-        if (!status) {
-          return false;
-        }
-
         if (area === 'main') {
-          openInMain(model);
+          // Temporarily bypass model disposal to transport model to main view
+          // to keep the conversation when switching views
+          const originalDispose = previousModel.dispose.bind(previousModel);
```
brichet, line 1099: I like the idea of keeping the model when moving the widget 👍 .

Overriding temporary the `dispose` method is a nice workaround for now.
I opened https://github.com/jupyterlab/jupyter-chat/pull/423 to fix it in `jupyter-chat` instead.

We should add a comment about removing this temporary code when bumping `jupyter-chat`.

brichet (commented):
Thanks, it works well.

Most of the comments from the previous review were not addressed (and not commented) so I unresolve them.

Yahiewi:
> Thanks, it works well.
> 
> Most of the comments from the previous review were not addressed (and not commented) so I unresolve them.

Hey @brichet thanks for rereviewing.
I'm sorry but I think I messed up the branch a bit because there seem to be local commits missing here and the remote and local branches have diverged. I'll work on fixing that and then address your new comments.

src/chat-model.ts
Yahiewi: Not really, I set it back to private.

src/widgets/main-area-chat.ts
Yahiewi: Done

src/index.ts
Yahiewi: Done

src/chat-model.ts
Yahiewi: Done

src/chat-model.ts
Yahiewi: Done

src/chat-model.ts
```diff
@@ -287,12 +288,27 @@ export class AIChatModel extends AbstractChatModel implements IAIChatModel {
...
     this.messagesDeleted(0, this.messages.length);
     this.title = null;
     this._toolContexts.clear();
     await this._agentManager.clearHistory();
   };
 
+  /**
+   * Overrides messageAdded to ensure queued messages stay at the bottom.
+   */
+  override messageAdded(message: IMessageContent): void {
```
Yahiewi, line 305: Before I added this, it used to be when you have queued messages and active tool calls, the queued message can lag behind in the chat UI, with the agent messages being below them. This ensure they remain at the bottom of the UI.

Yahiewi:
I merged the divergent branches and now all the old and new reviews should be properly addressed.
I also tested again, and everything seems to be working as expected now.

brichet:
@Yahiewi I pushed a small commit to address https://github.com/jupyterlite/ai/pull/331#discussion_r3155501212

brichet:
cc. @jtpio, in case you want to take a quick look at it.

Summary of the main changes:
- add messages to a queue if agent is busy (already streaming a message)
- display the queued messages in a temporary message (mime message) at the end of the chat. This queued message is reset when (1) the messages change in the queue or (2) a message is added to the chat (either the agent or a queued messages being processed), to keep it at the end of the chat
- new method `_process_message` in chat model that send the message to the agent. This method is called by
  - the `sendMessage` method (same behavior as before) 
  - the `_drainQueue` method, that will handle the next message in queue when the previous one is processed

src/chat-model.ts
```diff
@@ -413,6 +449,90 @@ export class AIChatModel extends AbstractChatModel implements IAIChatModel {
...
+      this.updateWriters([]);
+      this._removeQueueUI();
+      return;
+    }
+
+    // Dequeue and push to chat
+    const next = this._messageQueue.shift()!;
+    next._originalMsg.time = Date.now() / 1000;
+    this.messageAdded(next._originalMsg);
+    this._updateQueueUI();
```
```suggestion
```

I guess this is not necessary anymore, it should be handled in the overridden `messageAdded` method.
To be tested, though.

brichet (approved):
Thanks @Yahiewi, it looks good to me.

We should probably add tests on it in a follow up PR.

src/chat-model.ts
Yahiewi: Thanks for catching this, I removed this duplicated code and tested again and everything works as expected.

jtpio (approved):
Thanks, let's try it!

PR #335, opened 2026-04-28
Export an interface for the ai chat model

brichet:
This PR exports a new interface for the `AIChatModel` object.
Currently the `AIChatModel` was not exported, which result in some difficulties to work with.
For example, the chat tracker model could not be cast, therefore some features were not available (save, restore...) from a third party extension. This avoid casting the model as `any`.

### DIFF
--- a/src/tokens.ts
+++ b/src/tokens.ts
@@ -682,7 +682,7 @@ export interface IAIChatModel extends IChatModel {
   /**
    * A signal emitting when the chat name has changed.
    */
-  nameChanged: ISignal<IAIChatModel, string>;
+  readonly nameChanged: ISignal<IAIChatModel, string>;
   /**
    * The title of the chat.
    */
@@ -706,7 +706,7 @@ export interface IAIChatModel extends IChatModel {
   /**
    * A signal emitting when the token usage changed.
    */
-  tokenUsageChanged: ISignal<IAgentManager, ITokenUsage>;
+  readonly tokenUsageChanged: ISignal<IAgentManager, ITokenUsage>;
   /**
    * The agent manager used in the model.
    */
--- a/src/tokens.ts
+++ b/src/tokens.ts
@@ -1,4 +1,4 @@
-import { ActiveCellManager, IMessage } from '@jupyter/chat';
+import { ActiveCellManager, IChatModel, IMessage } from '@jupyter/chat';
 import { VDomRenderer } from '@jupyterlab/apputils';
 import { IRenderMimeRegistry } from '@jupyterlab/rendermime';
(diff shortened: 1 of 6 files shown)


src/widgets/main-area-chat.ts
```diff
@@ -78,14 +78,14 @@ export class MainAreaChat extends MainAreaWidget<ChatWidget> {
...
+    this.model.writersChanged?.disconnect(this._writersChanged);
     this.model.titleChanged.disconnect(this._titleChanged);
   }
 
   /**
    * Get the model of the chat.
    */
-  get model(): AIChatModel {
+  get model(): IAIChatModel {
     return this.content.model as AIChatModel;
```
jtpio: Do we still need the cast to `AIChatModel` here btw?

src/tokens.ts
```diff
@@ -665,14 +669,64 @@ export const IAgentManagerFactory = new Token<IAgentManagerFactory>(
 
 /* THE CHAT MODELS HANDLER */
 
+export interface IAIChatModel extends IChatModel {
+  /**
+   * A signal emitting when the chat name has changed.
+   */
+  nameChanged: ISignal<IAIChatModel, string>;
```
jtpio: Maybe this could be made `readonly`?


```suggestion
  readonly nameChanged: ISignal<IAIChatModel, string>;
```

src/tokens.ts
```diff
@@ -665,14 +669,64 @@ export const IAgentManagerFactory = new Token<IAgentManagerFactory>(
...
+   */
+  readonly autosaveChanged: ISignal<IAIChatModel, boolean>;
+  /**
+   * Whether save/restore is available.
+   */
+  readonly saveAvailable: boolean;
+  /**
+   * A signal emitting when the token usage changed.
+   */
+  tokenUsageChanged: ISignal<IAgentManager, ITokenUsage>;
```
jtpio: Maybe this one too could be made `readonly`?


```suggestion
  readonly tokenUsageChanged: ISignal<IAgentManager, ITokenUsage>;
```

jtpio (approved):
Thanks!

Left a couple of minor comments, otherwise looks good 👍

src/widgets/main-area-chat.ts
brichet: Yes, the default is `IChatModel`.
I'll update the cast to be `IAIChatModel` instead of `AIChatModel` for consistency.

brichet:
Thanks for the review @jtpio, I updated the PR with your suggestions.

jtpio (approved):
Thanks!

PR #336, opened 2026-04-28
Save the chat when autosave is updated

brichet:
Before this PR, the chat was not saved when autosaved was unset.
Since the autosave state is also saved in file, this led to inconsistencies when reloading the chat.


jtpio (approved):
Thanks!

PR #337, opened 2026-04-28
Title follow up

brichet:
This PR follows up #327 
- remove the previous title when the messages are cleared
- add a setting to automatically request a title for every message until 5
- enhance the prompt when requesting the title, to focus on subject instead of actions

### DIFF
--- a/src/chat-model.ts
+++ b/src/chat-model.ts
@@ -481,7 +481,7 @@ export class AIChatModel extends AbstractChatModel {
       {
         role: 'system',
         content:
-          "Generate a concise title (no more than 10 words) for the following conversation. Do not use formatting. Focus on the user's main intent."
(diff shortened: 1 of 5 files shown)


src/chat-model.ts
```diff
@@ -347,6 +348,17 @@ export class AIChatModel extends AbstractChatModel {
       this.updateWriters([{ user: this._getAIUser() }]);
 
       await this._agentManager.generateResponse(enhancedMessage);
+
+      if (
+        this._settingsModel.config.autoTitle &&
+        (this.messages.length <= 5 || this.title === null)
+      ) {
+        try {
+          this.title = await this.requestTitle();
```
jtpio: Wondering if we should clear the writers before requesting the title?

Because otherwise the chat UI may show the stop button, even though it's not meant to interrupt the title generation request?

jtpio (approved):
Thanks!

Just left a small question, but otherwise this looks good to go.

src/chat-model.ts
brichet: Good catch, thanks.

I moved the title request to the `finally` statement instead, the error should be caught anyway.

jtpio (approved):
Thanks!

PR #340, opened 2026-05-21
Title requested for the 5 firsts user messages

brichet:
A small fix when `autotitle` is enabled, to request title during the firsts 5 messages of the user (and not the firsts 5 messages in total).
This should help having a more relevant title on the chat.


jtpio (approved):
Thanks!

PR #344, opened 2026-06-03
Support optional shell commands

jtpio:
To leverage the new (optional) commands added in https://github.com/jupyterlite/terminal/pull/110

- [x] Include `jupyterlite-terminal` in the demo environment only
- [x] Update the system prompt to hint at optional shell commands
- [ ] Mention in the docs?

https://github.com/user-attachments/assets/49f2780c-d1c4-4aaf-84dd-78ad3ff38c0f


jtpio:
Opening early to test out the idea. Technically the "feature" could also be encoded as a skill without having to change the system prompt (trying to evaluate what the best balance should be).

PR #347, opened 2026-06-06
Enable prompt caching for Anthropic models

### LINKED ISSUE #332: Use prompt caching for Anthropic models
As mentioned in #312 we can use prompt caching for Anthropic models as explained here:
https://ai-sdk.dev/cookbook/node/dynamic-prompt-caching
Although, this is apparently only supported by Anthropic models, we can still implement it as "for non-Anthropic models, messages pass through unchanged, making this safe to use in provider-agnostic code."
Plus, I think Anthropic models are one of the most widely used (if not the most), so this would be a useful feature for a lot of people: "With Anthropic, cached tokens cost 10% of input tokens, but cache writes cost 25% more. You save money when cache hits exceed cache misses."
So while it might cost slightly more for very short conversations, it can save up A LOT of tokens for longer conversations.

MUFFANUJ:
Closes #332 

Adds Anthropic prompt caching by marking the latest conversation message with an ephemeral cache breakpoint. Messages for non-Anthropic models remain unchanged.

### DIFF
--- a/src/agent.ts
+++ b/src/agent.ts
@@ -898,6 +898,9 @@ ${richOutputWorkflowInstruction}`;
       tools,
       temperature,
       maxOutputTokens,
+      prepareStep: ({ messages, model }) => ({
+        messages: Private.addCacheControlToMessages(messages, model)
+      }),
       stopWhen: stepCountIs(maxTurns)
     });
   }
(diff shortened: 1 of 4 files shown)


src/agent.ts
```diff
@@ -1303,6 +1306,40 @@ WEB RETRIEVAL POLICY:
...
+    if (
+      messages.length === 0 ||
+      !(
+        (typeof model === 'string' &&
+          (model.includes('anthropic') || model.includes('claude'))) ||
+        (typeof model !== 'string' &&
+          (model.provider.includes('anthropic') ||
+            model.modelId.includes('anthropic') ||
+            model.modelId.includes('claude')))
+      )
```
brichet: I wonder if we could be more generic, by using [`IProviderInfo`](https://github.com/jupyterlite/ai/blob/d6a9c07fa818672e151dece65c04a1abf8ea82f3/src/tokens.ts#L226) instead of having the comparison in the function. Others providers may likely enable caching soon.
We could then add, for example, `cacheProviderOptions` [here](https://github.com/jupyterlite/ai/blob/d6a9c07fa818672e151dece65c04a1abf8ea82f3/src/providers/built-in-providers.ts#L14) containing `anthropic: { cacheControl: { type: 'ephemeral' } }`.

brichet (commented):
Thanks @MUFFANUJ for starting this, it works well overall.

Not necessarily for this PR, but we could also add the cache control to the tools, there is also a `providerOptions`: https://github.com/vercel/ai/blob/eeefc3f64920fc4f576263f1272194e004edae4d/packages/provider-utils/src/types/tool.ts#L73

brichet (approved):
Thanks @MUFFANUJ, LGTM.

PR #350, opened 2026-06-06
Make oversized chat prompts recoverable and add context usage warning

### LINKED ISSUE #291: Error when prompt has too many tokens
I sent the `1-api-key.png` under `./img/` twice to the Mistral model and it gives me another `Error generating response: No output generated. Check the stream for errors.` and in the console raises the following error:
`AI_APICallError: Prompt contains 152960 tokens and 0 draft tokens, too large for model with 131072 maximum context length`

Mistral (mistral-medium-latest):
Image

Claude Opus 4.6 (much larger context window at 1M tokens):
Image

This error as usual blocks the conversation and you need to start another conversation which isn't ideal.
It would be nice to make going over the context window limit a non-blocking error. 
Or perhaps instead of trying to fix this issue, if we had the option to go back to a message and restart the conversation from there after editing it, this can bypass the issue by giving users the option to revert the history to before that error and they would avoid going over the context window then.

MUFFANUJ:
Closes #291 

This change makes oversized provider prompt errors non-blocking by avoiding persistence of failed turns while still cleaning rejected attachment history. It also supports editing a previous user message to restart the conversation from that point, and adds a dismissible context warning above the chat input when the last request reaches 95% of the model context window.

### DIFF
--- a/src/agent.ts
+++ b/src/agent.ts
@@ -631,19 +631,13 @@ export class AgentManager implements IAgentManager {
       this._history.push(...Private.sanitizeModelMessages(responseHistory));
     } catch (error) {
       if ((error as Error).name !== 'AbortError') {
-        let helpMessage = `${(error as Error).message}`;
+        const payloadRejectionError = Private.getPayloadRejectionError(error);
+        let helpMessage = `${(payloadRejectionError ?? (error as Error)).message}`;
 
         // Remove attachments from history on payload rejection errors
-        if (
-          APICallError.isInstance(error) &&
-          (error.statusCode === 400 ||
-            error.statusCode === 404 ||
-            error.statusCode === 413 ||
-            error.statusCode === 415 ||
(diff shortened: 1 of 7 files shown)


MUFFANUJ:
Please ignore the context window warning i enabled it for testing purpose below is the updated one:

https://github.com/user-attachments/assets/10c359ab-120b-4156-b72d-c821719f9879

Screenshot 2026-06-07 at 2 09 19 AM

brichet (commented):
Thanks @MUFFANUJ for working on this.

This PR should probably be split, it seems to address 3 different issues/enhancements:
- fixing message history when an error occurs
- allowing to update a message and start a new conversation (also started at https://github.com/jupyterlite/ai/pull/310, IIUC)
- adding a context warning component - I wonder if we should mix it with the current context usage component in the toolbar https://github.com/jupyterlite/ai/pull/262

PR #353, opened 2026-06-11
Monorepo with agent and extension

brichet:
This PR splits the javascript package to a monorepo:
- a frontend agent, currently named ~`@jupyterlite/agent` (to be renamed)~ `@jupyternaut/agent`
- a lab extension, currently named `@jupyterlite/ai` (to be renamed too)

There is no change in the Python package in the initial state of the PR.

### Packaging changes
- a _packages_ directory containing the _agent_ and _ai_ workspaces.
- a main _package.json_ file responsible for handling the workspaces and the linting scripts
- a new _scripts/bump_version.py_ to handle version bump

### Discussion before undrafting the PR
- We probably need to rename of the project and packages in this PR
- The tokens relative to the agent are defined in ~`@jupyterlite/agent`~ `@jupyternaut/agent` package.
  I'm not sure that it makes sense, and maybe they should be defined in `@jupyterlite/ai` ?
  Each extension making use of the agent would then defined its own token.
- All the versions are the same (python package, `@jupyterlite/ai` and ~`@jupyterlite/agent`~ `@jupyternaut/agent`), `0.18.0.
  Should we keep the same version number for all ? Should we start again, if we rename the packages ?


packages/ai/package.json
```diff
@@ -0,0 +1,121 @@
...
+    "@lumino/messaging": "^2.0.3",
+    "@lumino/polling": "^2.1.4",
+    "@lumino/signaling": "^2.1.4",
+    "@lumino/widgets": "^2.7.1",
+    "@mui/icons-material": "^7",
+    "@mui/material": "^7",
+    "ai": "^6.0.116",
+    "jupyter-chat-components": "^0.6.0",
+    "jupyter-secrets-manager": "^0.5.0",
+    "yaml": "^2.8.1",
```
jtpio: Looks like the `yaml` dependency is used in the `agent` package, maybe we should move it there?

jtpio (approved):
Thanks!

Looks good overall (left a minor comment)

packages/ai/package.json
brichet: Good catch, thanks.
There were actually other dependencies to fix or homogenize between package, I updated it.

Issue #355, opened 2026-06-19
Cleaning the sources


brichet:
Some code in the source seems to be leftover and may be cleaned.

- I can't find reference to the command `'@jupyterlite/ai:reposition'` 
  https://github.com/jupyterlite/ai/blob/9ef5e207a02a2d25d22701fa20ab676ec6a87040/packages/ai/src/index.ts#L739

  It has been introduced in #145, which was a big refactor on the agent workflow, and I wonder if this is used somewhere.

- The diff manager, used to display cells and file diff after modifications, is probably useless since #219, which make use of `jupyterlab-ai-commands` to edit files and notebook, and which handles itself the diff if expected.
  https://github.com/jupyterlite/ai/blob/9ef5e207a02a2d25d22701fa20ab676ec6a87040/packages/ai/src/diff-manager.ts#L28
  https://github.com/jupyterlite/ai/blob/9ef5e207a02a2d25d22701fa20ab676ec6a87040/packages/ai/src/index.ts#L1438

cc. @jtpio, since you were involved in these changes, to confirm that they can be removed.

brichet:
> * I can't find reference to the command `'@jupyterlite/ai:reposition'`

I opened https://github.com/jupyter-ai-contrib/jupyterlab-ai-commands/pull/26 to move this command to `jupyterlab-ai-commands`

PR #356, opened 2026-06-24
Jupyternaut persona extension

brichet:
A new extension that exposes a frontend persona, that aims to be available for any extension using a chat.

### Goal

The goal of this change is to allow adding a frontend agent to the chats widgets tracked with the chat tracker.
It makes it compatible with `jupyter-ai`, and can be installed as a regular extension, without necessarily embed it own chat system.

It is a step forward for homogenizing `jupyter-ai` and `jupyterlite-ai`, and have a similar user experience in jupyterlab and jupyterlite.

### Major changes

1. The repository now provides 3 JavaScript packages:
  - `@jupyternaut/agent`, a package relying on `@ai-sdk` to build a frontend agent
  - `@jupyternaut/persona`, a package providing plugins to add a frontend persona
    - contains a class `Persona` that link a chat model to a `@jupyternaut/agent`, using the `IChatTracker`. By default it respond to message when mentioned (currently `@jupyternaut-frontend`, should be updated), but it can be set up to answer to every message
    - contains a class `PersonaRegistry` that map each chat model to its agent (available from a token)
    - if a chat is created without persona, one is automatically created and attached to it
    - provides all the tokens necessary for the agent (providers, settings...)
  - `@jupyterlite/ai` which creates the chat model and add components (button) to the chat widget. It also set up the persona to respond to every message from user, to keep the same UX as before

2. It also provides 2 jupyterlab extensions, one for the persona only, and one for the chat (that require the persona one).

3. The settings are split in 2 places
  - the settings related to the agent lands in a dedicated panel in `@jupyternaut/persona` package, available from the *settings* menu (and the `jupyter-ai` chat toolbar as before)
  - the settings related to chat (`sendWithShiftEnter`, `showTokenUsage`, ...) are in the regular Jupyterlab settings system 
 
### Discussion

- currently the agent is mentioned with `@jupyternaut-frontend` to avoid conflict with the server side  `@jupyternaut`. This mention should be renamed IMO (except if this is only a transitive name).

### Follow up

Currently the persona is not compatible with `jupyter-ai` because the persona can't propagate message in a chat from `jupyterlab-chat` (see https://github.com/jupyterlab/jupyter-chat/issues/454).
Maybe we can move forward with this PR and wait for https://github.com/jupyterlab/jupyter-chat/pull/455 to be release to update `@jupyter/chat` dependency and fix this issue.


packages/ai/package.json
```diff
@@ -101,10 +94,22 @@
...
       "@jupyter/chat": {
         "bundled": true,
         "singleton": true
       },
+      "@jupyternaut/agent": {
+        "bundled": false,
+        "singleton": true
+      },
+      "@jupyternaut/persona": {
+        "bundled": false,
```
jtpio, line 110: Should we have the `jupyterlite-ai` Python package depend on `jupyternaut-persona` if `@jupyternaut/persona` is expected to be on the page?

packages/ai/src/chat-model.ts
```diff
@@ -347,110 +307,56 @@ export class AIChatModel extends AbstractChatModel implements IAIChatModel {
...
-          this.title === null)
-      ) {
-        try {
-          this.title = await this.requestTitle();
-        } catch (e) {
-          console.warn('Error while generating a title\n', e);
-        }
+    if (
+      this._settings?.composite['autoTitle'] === true &&
+      (this.messages.filter(msg => msg.sender.username !== 'ai-assistant')
```
jtpio: Should it be `sender.bot` instead of `'ai-assistant'`?

packages/persona/src/components/completion-status.tsx
```diff
@@ -1,8 +1,8 @@
-import { useEffect, useState } from 'react';
 import { ReactWidget } from '@jupyterlab/ui-components';
 import type { TranslationBundle } from '@jupyterlab/translation';
 import { jupyternautIcon } from '@jupyternaut/agent';
-import type { IAISettingsModel } from '../tokens';
+import type { IAISettingsModel } from '@jupyternaut/agent';
+import React, { useEffect, useState } from 'react';
 
 const COMPLETION_STATUS_CLASS = 'jp-ai-completion-status';
 const COMPLETION_DISABLED_CLASS = 'jp-ai-completion-disabled';
```
jtpio, line 8: Looks like the corresponding CSS rules should be moved from the `ai` package to the `base.css` file in `jupyternaut-persona`?

packages/persona/schema/settings-model.json
jtpio, line 1: Wondering if this change should be documented in the changelog since this would likely break settings for existing installs?

This could be done during the release process though (editing the changelog for the GitHub draft release before starting step 2).

packages/persona/src/persona.ts
```diff
@@ -0,0 +1,622 @@
...
+  private _computeShouldAutoRenderMimeBundles(
+    toolName: string,
+    input: string
+  ): boolean {
+    if (toolName !== 'execute_command') {
+      return false;
+    }
+    try {
+      const parsed = JSON.parse(input);
+      console.log(
```
jtpio: Maybe to remove if it was for debugging?

packages/persona/src/index.ts
```diff
@@ -0,0 +1,703 @@
...
+  ): IAISettingsModel => {
+    return new AISettingsModel({ settingRegistry });
+  }
+};
+
+/**
+ * Diff manager plugin
+ */
+const diffManager: JupyterFrontEndPlugin<IDiffManager> = {
+  id: '@jupyterlite/ai:diff-manager',
```
jtpio: Should this be renamed for consistency?


```suggestion
  id: '@jupyternaut/persona:diff-manager',
```

packages/persona/src/tokens.ts
```diff
@@ -0,0 +1,121 @@
...
+  bot: true,
+  mention_name: 'jupyternaut-frontend'
+};
+
+/**
+ * Command IDs namespace
+ */
+export namespace CommandIds {
+  export const openSettings = '@jupyternaut/persona:open-settings';
+  export const refreshSkills = '@jupyternaut/persona:refresh-skills';
```
jtpio, line 27: Should the command id be updated in the agent package too?

https://github.com/jupyterlite/ai/blob/da63e32262d8c1a2305bde91910c07e09cfdd8af/packages/agent/src/tokens.ts#L21

jtpio (commented):
Thanks for tackling this refactoring! 

Left a couple of comments.

packages/persona/src/tokens.ts
brichet: The command ID `refreshSkills` should actually have been removed from the agent package.

packages/persona/src/index.ts
brichet: Right, but IMO this plugin should be removed https://github.com/jupyterlite/ai/issues/355

packages/ai/package.json
brichet: Isn't it required to have all the plugins from `jupyternaut-persona` activated ?

packages/persona/src/index.ts
brichet: Updated in https://github.com/jupyterlite/ai/pull/356/commits/db7343bbd7b1c3b63aaf43e8b579b31936709b4f.
Let's do the removal (if expected) in a follow up PR.

packages/persona/src/tokens.ts
brichet: Done in https://github.com/jupyterlite/ai/pull/356/commits/db7343bbd7b1c3b63aaf43e8b579b31936709b4f

packages/persona/schema/settings-model.json
brichet: Right, we should update it in the draft release.

jtpio (approved):
Thanks!

Let's get this in so it's easier to iterate.

PR #357, opened 2026-07-30
Fix default system prompt to prevent stopping before tool calls

jtpio:
Seems like the agent may sometimes get stuck, at least when trying with Mistral as the provider:

**Before**

jupyterlite-ai-stuck-before-fix

**After**

jupyterlite-ai-tool-calls-after-fix


brichet (approved):
Thanks @jtpio, LGTM.

As you suggested in another channel we should probably reduce this system prompt at some point, most of the model may work fine without it. We should find a way to compare the results according to the system prompt for different model, even if it is ran manually (to use API key).

PR #358, opened 2026-07-30
Update to AI SDK v7

jtpio:
Reference: https://ai-sdk.dev/docs/migration-guides/migration-guide-7-0

- [x] Update dependencies
- [x] Adapt to some renames and deprecation notices
- [x] Update some other dev dependencies

### DIFF
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -228,7 +228,11 @@ export class AgentManagerFactory implements IAgentManagerFactory {
         const client = await createMCPClient({
           transport: {
             type: 'http',
-            url: serverConfig.url
+            url: serverConfig.url,
+            // The transport calls this option as a method (`this.fetchFn(...)`),
+            // so an unbound window.fetch would throw "Illegal invocation" in
+            // browsers.
+            fetch: globalThis.fetch.bind(globalThis)
           }
         });
 
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -229,6 +229,9 @@ export class AgentManagerFactory implements IAgentManagerFactory {
           transport: {
             type: 'http',
             url: serverConfig.url,
+            // The transport calls this option as a method (`this.fetchFn(...)`),
+            // so an unbound window.fetch would throw "Illegal invocation" in
+            // browsers.
             fetch: globalThis.fetch.bind(globalThis)
           }
         });
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -228,7 +228,8 @@ export class AgentManagerFactory implements IAgentManagerFactory {
         const client = await createMCPClient({
           transport: {
             type: 'http',
-            url: serverConfig.url
+            url: serverConfig.url,
+            fetch: globalThis.fetch.bind(globalThis)
           }
         });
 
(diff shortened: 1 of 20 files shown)


packages/agent/src/agent.ts
```diff
@@ -220,7 +228,8 @@ export class AgentManagerFactory implements IAgentManagerFactory {
         const client = await createMCPClient({
           transport: {
             type: 'http',
-            url: serverConfig.url
+            url: serverConfig.url,
+            fetch: globalThis.fetch.bind(globalThis)
```
jtpio: Looks like this may be a regression in `@ai-sdk/mcp` v2, but maybe we can live with it for now.

packages/agent/src/agent.ts
```diff
@@ -48,6 +49,13 @@ interface IMCPClientWrapper {
   client: MCPClient;
 }
 
+/**
+ * The stream result type produced by the agent.
+ */
+type AgentStreamResult = Awaited<
```
brichet, line 55: Isn't there any type defined in `ai-sdk` that would natively handle the stream result ?

packages/agent/src/tools/commands.ts
```diff
@@ -95,7 +95,6 @@ function searchCommands(
  */
 export function createDiscoverCommandsTool(commands: CommandRegistry): ITool {
   return tool({
-    title: 'Discover Commands',
```
brichet, line 98: Should we still provide the title with the `providerMetadata` argument ? Or is this useless ?

packages/agent/src/agent.ts
brichet: Strangely, the default seems to already be the `globalThis.fetch`, according to the documentation.

brichet (approved):
Thanks @jtpio for working on this. Works well on my side, I just have some minor comments.

It seems that it also fix some errors, like the system message and token count in the `generateText` method of the agent :+1:

packages/agent/src/agent.ts
jtpio: The native type seems to be `StreamTextResult<ToolMap, Context, never>` but `Context` is not exported. So we would need to copy that type here too which may make it drift over time.

packages/agent/src/tools/commands.ts
jtpio: Looks like `title` is now deprecated and the tools calls were not using it anyway (showing `discover_commands` instead).

But maybe `metadata` can work.

packages/agent/src/agent.ts
jtpio: There seems to be a difference depending on whether this runs in node or in the browser. Will try to push a few commits to the branch to double check that.

packages/agent/src/tools/commands.ts
jtpio: Using `metadata`:

image

jtpio:
https://github.com/jupyterlite/ai/pull/358/changes/37df312439ec1ee9f5efd7fc9447253e55ec4ed9 without the fetch bind fix gives the following in the UI tests:

```
TypeError: Failed to execute 'fetch' on 'Window': Illegal invocation
```

The following commit passes https://github.com/jupyterlite/ai/pull/358/commits/cfc184cb7118e117f21b4385011a05cccb7439e4. Probably an issue we could report in the ai-sdk repo at some point? But for now probably it's fine to keep that around?

brichet:
Thanks @jtpio, it looks good to me.

PR #359, opened 2026-07-31
Bump jupyter chat 0.24

brichet:
This PR bumps `@jupyter/chat` to `0.24`

### DIFF
--- a/packages/ai/src/widgets/main-area-chat.ts
+++ b/packages/ai/src/widgets/main-area-chat.ts
@@ -1,20 +1,17 @@
 import { ChatArea, ChatWidget, IChatModel, IChatPanel } from '@jupyter/chat';
-import { CommandToolbarButton, MainAreaWidget } from '@jupyterlab/apputils';
-import type { TranslationBundle } from '@jupyterlab/translation';
-import { ISettingRegistry } from '@jupyterlab/settingregistry';
-import { launchIcon } from '@jupyterlab/ui-components';
-import { CommandRegistry } from '@lumino/commands';
+import { MainAreaWidget } from '@jupyterlab/apputils';
 
-import { SaveComponentWidget } from '../components/save-button';
-import { UsageWidget } from '../components/usage-display';
+// import { SaveComponentWidget } from '../components/save-button';
+// import { UsageWidget } from '../components/usage-display';
 import { RenderedMessageOutputAreaCompat } from '../rendered-message-outputarea';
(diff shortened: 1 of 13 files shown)


brichet:
> Currently the persona is not compatible with jupyter-ai because the persona can't propagate message in a chat from jupyterlab-chat (see https://github.com/jupyterlab/jupyter-chat/issues/454).

_Commented by @brichet in https://github.com/jupyterlite/ai/pull/356#issue-4734794456_

While https://github.com/jupyterlab/jupyter-chat/pull/455 has been merged, this is not enough to make the persona extension compatible with jupyter-ai, because we still need the message ID when a bot message has been sent. It would allow replacing the following `addMessage` with `sendmessage`, and store the streaming message.
https://github.com/jupyterlite/ai/blob/46bae596a5cc98929cb6e6dc1ac886f1f3f45f95/packages/persona/src/persona.ts#L372-L374

https://github.com/jupyterlab/jupyter-chat/pull/477 should help with it, by return the message ID when a message is sent.

packages/ai/src/widgets/main-area-chat.ts
```diff
@@ -1,72 +1,55 @@
...
-import { ISettingRegistry } from '@jupyterlab/settingregistry';
-import { launchIcon } from '@jupyterlab/ui-components';
-import { CommandRegistry } from '@lumino/commands';
+import { ChatArea, ChatWidget, IChatModel, IChatPanel } from '@jupyter/chat';
+import { MainAreaWidget } from '@jupyterlab/apputils';
 
-import { SaveComponentWidget } from '../components/save-button';
-import { UsageWidget } from '../components/usage-display';
+// import { SaveComponentWidget } from '../components/save-button';
+// import { UsageWidget } from '../components/usage-display';
```
jtpio: Should these be removed?

packages/persona/src/persona.ts
```diff
@@ -25,7 +26,7 @@ import {
   modelSupportsPdf
 } from '@jupyternaut/agent';
 
-import { UUID } from '@lumino/coreutils';
+// import { UUID } from '@lumino/coreutils';
```
jtpio: This one (and above) could be removed too?

jtpio (approved):
Thanks!

Some commented code could likely be removed, otherwise looks good to go.

brichet:
Thanks for the review @jtpio

PR #363, opened 2026-08-18
Jupyternaut compatibility with jupyter-ai 3.1.3

brichet:
Register a frontend persona in jupyter-ai persona manager, using the `jupyter_ai.personas` entry point.

The back end part of the persona does nothing on message, because the frontend persona handles it directly.

Some changes are expected on `jupyter-chat`:
- https://github.com/jupyterlab/jupyter-chat/pull/500 (backported to 0.24.x in https://github.com/jupyterlab/jupyter-chat/pull/502)
- https://github.com/jupyterlab/jupyter-chat/pull/503 (will be backported in https://github.com/jupyterlab/jupyter-chat/pull/504)
- https://github.com/jupyterlab/jupyter-chat/pull/505

## Main changes

- add a dependency to `jupyter-ai-persona-manager` to be able to create a persona in `jupyter-ai`. Unfortunately it brings a lot of unexpected dependencies: `jupyter-ai-router`, `jupyterlab-chat`, jupyter-collaboration`
- add a persona in the backend, that is added to the persona manager (but does nothing on new message, since the agent is in the frontend)
  - this backend persona needs to read the user settings to retrieve the list of available models
- `@jupyternaut-persona` read also the message metadata (filled by the `jupyter-ai` provider/model selector widget) to choose which provider to use, and rebuilt the agent if necessary
- the callback to update `groupedToolCall` component has been moved to the persona instead of `@jupyterlite/ai`
- same for the skills command, it has been moved to the persona
- messages from the agent are now sent (using `sendfMessage` instead of `messageInserted`) to propagate them to the backend if there is one
- `jupyterlab` and `jupyterlite-core` have been updated in the demo lock file (`4.6.3` and `0.8.3` respectively)


brichet:
It need to wait for a backport of https://github.com/jupyterlab/jupyter-chat/pull/532 in 0.24.x branch of jupyter-chat.
Otherwise, the chat may be set as ready before the messages are loaded, and the persona responds to previous messages.

brichet:
The PR should be ready, jupyternaut frontend is now compatible with `jupyter-ai 3.1`.

[record-2026-08-31_16.51.54.webm](https://github.com/user-attachments/assets/430617fe-cb51-49f5-bf98-dfb4e9f86036)


Currently the demo deployment is broken on main (https://github.com/jupyterlite/ai/issues/366), we should wait for https://github.com/jupyterlite/ai/pull/367 before merging, to ensure that everything works well in the deployment in documentation.

brichet:
#367 has been merged.

Another change to notice is that the queue component doesn't work when using `jupyternaut-persona` with `jupyter-ai`.
`jupyternaut-persona` listens for new messages in a chat to respond, while the queue component need to be inserted before the message is sent.
A hacky way to fix it could be to remove the message after it has been sent (when jupyternaut catch it), and add it to a queue component.

brichet:
Putting this PR on draft, some improvement on related projects will simplify a lot this change, to have compatibility:
- Jupyterlab-chat 0.25.0 is now compatible with Jupyterlite. See https://github.com/jupyterlab/jupyter-chat/pull/529
- https://github.com/jupyter-ai-contrib/jupyter-ai-persona-manager/pull/154 allows using the persona manager without backend

PR #364, opened 2026-08-18
Remove the message output rerender

brichet:
This PR removes the message output rerender, used to add a `jp-OutpuArea` class to each message.

This workaround was necessary to render cell output widgets, and was introduced in #272.

It still needs https://github.com/jupyterlab/jupyter-chat/pull/507 to be merged and released.


jtpio (approved):
Thanks!

PR #371, opened 2026-09-18
Expose `additionalInstructions`, fix tool call handling

jtpio:
Expose `additionalInstructions` to make it easier to extend the system prompt when the agent may be running elsewhere (for example in the terminal).

Also fix handling of parallel tool calls when more than one needs approval.


brichet (approved):
Thanks @jtpio, LGTM

PR #375, opened 2026-09-18
update the uv lock file in demo with latest jupyternaut-persona version

brichet:
This fix the version of the jupyternaut_persona in the demo.

I haven't check why it did not get updated in the latest release yet.


jtpio (approved):
Thanks!

PR #367, opened 2026-08-28
Add jupyternaut-persona as dependency of jupyterlite-ai

brichet:
Fixes #366

- updates the `jupyter-builder` dependency to `>=1.2.2` to fix installation with jupyterlab `<4.6`
- adds `jupyternaut-persona` as a dependency of `jupyterlite-ai` (an oversight of #356)
- fixes the tests and bump script related to this new dependency (internal to the project)
- add a test on the lite demo, to ensure that the extension is loaded

PR #378, opened 2026-09-22
Bump jupyternaut-persona in demo lock file

brichet:
Follow up of #367, which pinned the version of `jupyternaut-persona` in *uv.lock*.

The issue has been found after the first release including #367 (see #375)


jtpio (approved):
Thanks!

Looks like the `uv` version is now specified in three different places, which means they can start to drift away at some point. Maybe we can try to find a way to centralize that version number at some point.

